PR #2, opened 2020-10-21
Wffhcohort 80 v4

shelleyMurphy:
Tests added. I moved some common methods into another class and kept these tests separate from the CqlEngineWrapper tests, let me know if this isn't what you had in mind or if you prefer I move everything back into one class. 

Note: The test failure seems to be a classic case of "but it works locally", so I'm working on figuring out what could be causing that.


csandersdev (commented):
You and I hit a similar problem at about the same time with slightly different approaches to solving it. If you look at my PR #1, I moved the common test logic into a base class and made the CqlEngineWrapperTests inherit from that. I have a slight preference for my approach because I was able to move the WireMockRule into the parent and also added some common behavior there to deal with potential port conflicts on the build machine, but either approach is fine. One of us is going to go first and the other is going to have to deal with the aftermath.

Related to the test failures, that was absolutely a bug in the tests. I've got it fixed on my branch.

cohort-engine/src/test/java/com/ibm/cohort/engine/CqlTemporalTests.java
```diff
@@ -377,4 +378,25 @@ public Encounter getEncounter(int year, int month, int day) {
 		encounter.setPeriod(encounterPeriod);
 		return encounter;
 	}
+
+	private CqlEngineWrapper setupTestFor(Patient patient, FhirServerConfig fhirConfig, String... elm)
```
csandersdev, line 382: I'd like to see you move the setupTestFor(Patient...) methods from CqlEngineWrapper into BaseFhirTest since you need them vs. copy/pasting.

shelleyMurphy: I was going back and forth on that one, since I feel like that setup might be different as we add different tests, but I can change.

csandersdev (approved):
Thanks. If we find it doesn't make sense, we can always move it back out. Looks good to me now.

PR #3, opened 2020-10-21
Setup license header for VS Code users

deekim:
leveraging the [ymotongpoo/vsc-licenser](https://github.com/ymotongpoo/vsc-licenser) extension.

I wasn't sure if we wanted this to be committed since it is `IDE` specific,
but I thought that I would at least open up the discussion.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "licenser.license": "Custom",
+  "licenser.customHeader": "\n(C) Copyright IBM Corp. @YEAR@, @YEAR@\n\nSPDX-License-Identifier: Apache-2.0",
+  "licenser.useSingleLineStyle": false
+}
+


msargentibm:
Is this something you are using and sharing just in case others are interested?

deekim:
> Is this something you are using and sharing just in case others are interested?

It is what VS Code considers a ["workspace settings override"](https://code.visualstudio.com/docs/getstarted/settings)
> Workspace settings override user settings. 
> Workspace settings are specific to a project and can be shared across developers on a project.

csandersdev (approved):
Doesn't seem like it would interfere with another IDE, so I don't have a problem with it.

msargentibm:
>Doesn't seem like it would interfere with another IDE, so I don't have a problem with it.

Ditto. Were there any specific issues you were hoping to discuss @deekim ?

deekim:
> > Doesn't seem like it would interfere with another IDE, so I don't have a problem with it.
> 
> Ditto. Were there any specific issues you were hoping to discuss @deekim ?

Nope :)

PR #4, opened 2020-10-23
Thicks separate tests

ToddHicks:


csandersdev (changes requested):
You are bit out on your own right now because we don't know how this project will be consumed, but your project structure with engine, configfiles, and testObjects at the root is weird.

* I would recommend putting your Test.py under src/scripts or src/main/python
* I would rename configfiles to just config (mostly because that is what we did in previous WH projects) 
* I would put testObjects under src/main/resources and I would make the naming consistent with other folders (e.g. use camel case or not, but don't mix them)

ToddHicks:
> You are bit out on your own right now because we don't know how this project will be consumed, but your project structure with engine, configfiles, and testObjects at the root is weird.
> 
> * I would recommend putting your Test.py under src/scripts or src/main/python
> * I would rename configfiles to just config (mostly because that is what we did in previous WH projects)
> * I would put testObjects under src/main/resources and I would make the naming consistent with other folders (e.g. use camel case or not, but don't mix them)

I have made the changes. I renamed configfiles to scenarios, since that is what these files actually represent. I otherwise made the changes as you recommended.

csandersdev (approved):
Looks good.

PR #5, opened 2020-10-27
Refactor MeasureEvaluationSeed to be more configurable

deekim:
a first pass at a half-measure in making `MeasureEvaluationSeed` class more configurable

---

https://jsw.ibm.com/browse/WFFHCOHORT-149

### DIFF
--- a/cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
+++ b/cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
@@ -14,7 +14,9 @@ import org.hl7.fhir.r4.model.MeasureReport;
 import org.opencds.cqf.common.evaluation.EvaluationProviderFactory;
 import org.opencds.cqf.common.providers.LibraryResolutionProvider;
 import org.opencds.cqf.cql.engine.execution.LibraryLoader;
-import org.opencds.cqf.r4.evaluation.MeasureEvaluationSeed;
+
+import com.ibm.cohort.engine.measure.seed.IMeasureEvaluationSeed;
+import com.ibm.cohort.engine.measure.seed.MeasureEvaluationSeeder;
 
 import ca.uhn.fhir.rest.client.api.IGenericClient;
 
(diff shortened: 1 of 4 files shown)


cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
```diff
@@ -0,0 +1,175 @@
...
+
+		for (Map.Entry<Triple<String, String, String>, DataProvider> dataProviderEntry : dataProviders.entrySet()) {
+			Triple<String, String, String> usingDef = dataProviderEntry.getKey();
+			DataProvider dataProvider = dataProviderEntry.getValue();
+
+			context.registerDataProvider(usingDef.getRight(), dataProvider);
+		}
+
+		// todo: [daniel.kim] how is this different from the MeasurePeriod field?
+		context.setParameter(null, "Measurement Period",
```
csandersdev, line 117: I would like the name of this to default to "Measurement Period", but be configurable by the client. I'm going to push to keep the "standard" name, but I'm not sure we can 100% rely on it.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
```diff
@@ -0,0 +1,175 @@
...
+
+
+		TerminologyProvider terminologyProvider = createTerminologyProvider(usingDefs);
+		LinkedHashMap<Triple<String, String, String>, DataProvider> dataProviders = createDataProviders(usingDefs, terminologyProvider);
+		Interval measurementPeriod = createMeasurePeriod(periodStart, periodEnd);
+		Context context = createContext(library, dataProviders, terminologyProvider, measurementPeriod, productLine);
+
+		// todo: [daniel.kim] this only returns the last seen data provider
+		// is this a bug?
+		List<Map.Entry<Triple<String, String, String>, DataProvider>> dataProviderList = new ArrayList<>(dataProviders.entrySet());
```
csandersdev, line 89: The data provider that is stored in a member field and used downstream is used only once and that is to resolve the Patient class by patient ID. This _could_ go wrong if you had "using FHIR" and then later "using QDM" and the patient retrieval had to be FHIR. In reality, this is never going to happen. It would be a very strange use case to mix data providers. I think this is fine for now.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
csandersdev: I agree that the recreation of the interval object is strange. The only thing I can think of is that they wanted the context to have its own private copy of the variable so that if the engine messed with it, it wouldn't get change the original value in MeasurementSeed.

deekim:
After discussing this with Corey "offline," we agreed that if the `cql-ruler` maintainers were amenable, 
we would try to contribute this code back,  as we would have to maintain a copy of the code 
to be able to configure the class the way that we would want.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
```diff
@@ -72,14 +74,16 @@ public MeasureReport evaluatePatientMeasure(Measure measure, String patientId, S
...
+		seeder.disableDebugLogging();
 
 		// TODO - consider talking with OSS project about making source, user, and pass
 		// a properties collection for more versatile configuration of the underlying
 		// providers. For example, we need an additional custom HTTP header for
 		// configuration of our FHIR server.
-		seed.setup(measure, periodStart, periodEnd, /* productLine= */"ProductLine", /* source= */"", /* user= */"",
-				/* pass= */"");
+		IMeasureEvaluationSeed seed = seeder.create(measure, periodStart, periodEnd, "ProductLine");
+//		seed.setup(measure, periodStart, periodEnd, /* productLine= */"ProductLine", /* source= */"", /* user= */"",
```
jillrdoty, line 85: Can we get rid of this commented out code or add another comment about why it might be used in the future?

PR #7, opened 2020-10-28
Misc POM improvements

dkwasny-ibm:
* Adding configuration for the flattener plugin
* Moving `dependencyManagement` info from `cohort-engine/pom.xml` to the parent pom
* Adding the `<scm>` block to all poms

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,6 @@
 	
 	<modelVersion>4.0.0</modelVersion>
 
-	<!-- This shared pom should be flattened away during the build -->
 	<groupId>com.ibm.cohort</groupId>
 	<artifactId>cohort-parent</artifactId>
 	<version>${revision}</version>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -73,6 +71,117 @@
 		</repository>		
 	</repositories>
 
+	<dependencyManagement>
```
jillrdoty, line 74: Is there a reason for moving these dependencies?  My initial thought on leaving them was that they weren't necessarily needed by every sub-project and shouldn't be included if not.

dkwasny-ibm: This is the `<dependencyManagement>` block, which is effectively the "BOM" of the project.  Corey set up these versions to line up with the CQL engine and its dependencies back when `cohort-engine` was the only project.  These shouldn't force any dependencies on any subprojects, just the versions.

jillrdoty: It feels a little weird to call out the versions specifically above *and* have the dependencyManagement block... but I guess I don't care that much.

jillrdoty: Kwas explained that the versions are used for multiple dependencies, so I take it back.

jillrdoty (approved):

pom.xml
csandersdev: Yep. This is very important. I already ran into a case where cql-ruler was pulling in a different version of the engine than we were intending to use. Dependency management smooths it all out.

csandersdev (approved):

pom.xml
msargentibm: Thanks for the discussion. Had a similar train of thought as Jill, but this cleared things up.

msargentibm (approved):

PR #9, opened 2020-10-30
Add support for IBM CDM Measure resource with care-gap extensions

csandersdev:
The IBM Common Data Model FHIR IG defines an additional population type of care-gap that is not calculated by the underlying cqf-ruler implementation. I added a wrapper around MeasureEvaluation called CDMMeasureEvaluation that will pass through the bulk of the heavy lifting to cqf-ruler and then additionally calculate care-gap populations as needed.

While testing this work, I noted that the Library loading mechanism provided by cqf-ruler onlys supported CQL library attachments and not ELM. I added a custom Library loader that will handle both. Testing that uncovered that the library loading was quite chatty with four or more calls to FHIR for a single measure with a single Library. I added caching.

While working with QA it was noted that our docs around SSL were wrong, so I fixed that config and renamed the trustStore we provide for IBM Cloud CDT access to properly reflect its pkcs12 contents. I also added a help parameter to the command line options so that usage can be clearly viewed by a user who just has the JAR.

Some work was done to understand and verify the functionality of the CQL Engine when the provided ELM has been compiled using the CqlTranslator.Option.EnableDateRangeOptimization flag. This flag is necessary to support date filtering on the FHIR REST call, but it creates a scenario where the engine will misbehave when the FHIR retrieve provider does not understand the date property that is being used. I kept the test in place, but it is recommend *not* to enable that optimation flag at this time.


cohort-engine/.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,5 @@
+eclipse.preferences.version=1
```
jillrdoty, line 1: slight preference to add eclipse files to gitignore

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -0,0 +1,197 @@
...
+	 * 
+	 * @param measure   CDM Patient Quality Measure
+	 * @param context   CQL Engine Execution Context pre-configured for use in
+	 *                  measure evaluation
+	 * @param patientId Patient ID of the patient to evaluate
+	 * @return MeasureReport with population components filled out.
+	 */
+	public MeasureReport evaluatePatientMeasure(Measure measure, Context context, String patientId) {
+		MeasureReport report = evaluation.evaluatePatientMeasure(measure, context, patientId);
+		;
```
jillrdoty, line 107: unnecessary ;

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -0,0 +1,197 @@
...
+
+		public boolean inDenominatorExclusion() {
+			return inPopulation(MeasurePopulationType.DENOMINATOREXCLUSION);
+		}
+
+		public boolean inDenominatorException() {
+			return inPopulation(MeasurePopulationType.DENOMINATOREXCEPTION);
+		}
+
+		public boolean inNumerator() {
```
jillrdoty, line 56: Not sure if these method names reflect anything in cqf-ruler - if not, slight preference for renaming to something like "meetsNumeratorCriteria" or anything to reflect the fact that we're not trying to do the measure logic here (the patient is not necessarily in the numerator without evaluating the other populations).  

If I'm the only one who reads it that way then feel free to ignore, but I did spend an extra few minutes looking into this class trying to figure out if those methods were used to determine population results (which it seems not).

cohort-engine/src/test/resources/cql/valueset-override/test-with-vsparam.cql
```diff
@@ -0,0 +1,11 @@
+library "Test" version '1.0.0'
```
jillrdoty, line 1: Was this CQL your test to prove value set overrides don't work? Should it be included?

cohort-engine/src/test/resources/cql/includes/Breast-Cancer-Screening.cql
```diff
@@ -19,14 +19,7 @@ define "MeetsInclusionCriteria":
...
 define "MeetsExclusionCriteria":
-  "Breast Cancer"
-  or "Bilateral Mastectomy"
-  or "Unilateral mastectomy left and right."
-  or "Recent Mammogram"
-
-define "Unilateral mastectomy left and right.":
-  "Unilateral Mastectomy Left"
-  and "Unilateral Mastectomy Right"
+  false
```
jillrdoty, line 22: Did you mean to change this? I see the dummy CQLs so wasn't sure.

cohort-engine/src/test/resources/cql/valueset-override/test-with-vsparam.cql
csandersdev: Yes. Exactly as you said. It shouldn't have hit my branch. I'll remove it.

cohort-engine/src/test/resources/cql/includes/Breast-Cancer-Screening.cql
csandersdev: Intentional. This was failing to compile and I never noticed it. I recently added a check for translation errors and this popped out, so I had to fix it.

cohort-engine/.settings/org.eclipse.core.resources.prefs
csandersdev: I had been ignoring these files manually for a long time and then I finally gave in and committed them. I don't care either way. Would be happy to .gitignore if that is the group preference.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
csandersdev: Will fix.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
csandersdev: These are entirely my construction. I let the cqf-ruler do its thing for the standard populations and then I need to know after that runs whether or not a care-gap should be evaluated. The helper class checks the count for each population and then keeps track using a boolean whether they are in that population or not. I'm not sure "meets criteria" matches those semantics. For instance, if a patient meets both the numerator criteria and the numerator exception criteria, the are removed from the numerator (inNumerator = false and inNumeratorException = true). I'm not firm on this, so if you still think it makes sense to change the names I can do so.

cohort-engine/.settings/org.eclipse.core.resources.prefs
dkwasny-ibm: Ideally, all IDE specific files would be git ignored.  Our IDEs should be able to generate them from the `pom.xml` file no problem, and it prevents extra noise in further PRs as the IDEs change those metadata files.
My usual `.gitignore` entries for Eclipse is as follows.
```
bin
.classpath
.project
.settings
```

cohort-engine/src/test/java/com/ibm/cohort/engine/BaseFhirTest.java
```diff
@@ -170,18 +185,53 @@ protected CapabilityStatement getCapabilityStatement() {
...
-	protected CqlEngineWrapper setupTestFor(Patient patient, FhirServerConfig fhirConfig, String... elm)
+	protected CqlEngineWrapper setupTestFor(Patient patient, FhirServerConfig fhirConfig, String... resources)
 			throws Exception {
 
 		mockFhirResourceRetrieval("/metadata", getCapabilityStatement());
 		mockFhirResourceRetrieval(patient);
 
+		// This is a hack to get a single CQL resource from the provided set
+		// of library resources. This will be translated as needed.
+		final AtomicReference<String> cql = new AtomicReference<>();
```
dkwasny-ibm, line 208: Pedantic: Is there any reason to use `AtomicReference` here?  Are we doing some form of multithreading in our tests?

dkwasny-ibm (changes requested):
Anything marked with `Pedantic` may be ignored.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -0,0 +1,197 @@
...
+		public boolean inNumeratorExclusion() {
+			return inPopulation(MeasurePopulationType.NUMERATOREXCLUSION);
+		}
+
+		protected boolean inPopulation(MeasurePopulationType type) {
+			Boolean b = get(type);
+			if (b != null) {
+				return b.booleanValue();
+			} else {
+				return false;
```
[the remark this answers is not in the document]
csandersdev: The cqf-ruler does the calculations of the population types. The CDMMeasureEvaluation code runs after cqf-ruler and just needs to know if the patient is in or out of the different population groups. If the population group was not defined in the measure, the patient cannot belong, so I default to false. This only matters at the point where this class checks to see if the care gaps should be evaluated or not. I was not attempting to validate that the measure was defined appropriately. That would typically be done by the FHIR server at definition using the FHIR validation functionality and based in the IG definition.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -0,0 +1,197 @@
...
+		MeasureReport report = evaluation.evaluatePatientMeasure(measure, context, patientId);
+		;
+
+		MeasureScoring scoring = MeasureScoring.fromCode(measure.getScoring().getCodingFirstRep().getCode());
+		switch (scoring) {
+		case PROPORTION:
+		case RATIO:
+			// implement custom logic for CDM care-gaps
+			Iterator<Measure.MeasureGroupComponent> it = measure.getGroup().iterator();
+			for (int i = 0; it.hasNext(); i++) {
```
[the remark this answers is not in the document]
csandersdev: See line 117. I am using the loop index to dereference the report group. Could it have been written without combining the iterator and loop counter techniques? Maybe. I don't recall all the logic exactly, but I'm sure it made sense to me at the time.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -0,0 +1,197 @@
...
+		// each evaluation. That seems to be generally a bad thing unless
+		// the population components are coming from different libraries with
+		// potentially overlapping define names, but we _know_ that isn't happening
+		// because we are following the Davinci spec where there is only one Library
+		// entry point.
+
+		Object result = context.resolveExpressionRef(expression).evaluate(context);
+		if (result == null) {
+			result = Collections.emptyList();
+		}
```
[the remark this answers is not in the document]
csandersdev: This is based off the template in the cqf-ruler MeasureEvaluation class which does the thing with the emptyList.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/LibraryLoader.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.cqframework.cql.elm.execution.Library;
+import org.cqframework.cql.elm.execution.VersionedIdentifier;
+import org.hl7.fhir.r4.model.Attachment;
+import org.opencds.cqf.common.providers.LibraryResolutionProvider;
+import org.opencds.cqf.cql.engine.execution.CqlLibraryReader;
+
+import com.helger.commons.base64.Base64;
```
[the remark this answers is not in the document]
csandersdev: No. Stupid import resolver in Eclipse. Will fix.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/LibraryLoader.java
```diff
@@ -0,0 +1,125 @@
...
+ * CQL (text/cql) attachment. The ELM is preferred. When only CQL is present
+ * the loader will automatically translate to ELM using the configured
+ * translation provider
+ */
+public class LibraryLoader implements org.opencds.cqf.cql.engine.execution.LibraryLoader {
+
+	private Map<String, Library> libraries = new HashMap<>();
+
+	private LibraryResolutionProvider<org.hl7.fhir.r4.model.Library> provider;
+	private CqlTranslationProvider translationProvider;
```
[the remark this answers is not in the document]
csandersdev: The library cache? Certainly possible. The other two fields, I don't think so.

cohort-engine/src/test/java/com/ibm/cohort/engine/BaseFhirTest.java
csandersdev: AtomicReference is used as a container for a String so that it can be final, but the contained object can change. This enables use inside nested classes.

cohort-engine/src/main/java/com/ibm/cohort/engine/ClasspathLibrarySourceProvider.java
```diff
@@ -0,0 +1,29 @@
+package com.ibm.cohort.engine;
```
jillrdoty, line 1: No copyright?

cohort-engine/src/main/java/com/ibm/cohort/engine/EvaluationResultCallback.java
```diff
@@ -1,15 +1,6 @@
-/*
```
jillrdoty, line 1: Why remove copyright?

csandersdev: Good point and not the only one missing. I'll do a pass to add copyrights (again). Not sure why my Eclipse templating isn't working.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
jillrdoty: Hmm ok - I didn't read the inNumerator() as returning false if inNumeratorException() was true, but I understand better now.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -0,0 +1,197 @@
...
+	 * @return true when care gaps should be evaluated, otherwise false.
+	 */
+	private boolean isEligibleForCareGapEvaluation(MeasureReport.MeasureReportGroupComponent reportGroup) {
+		boolean isEligibleForCareGap = false;
+		StandardReportResults results = StandardReportResults.fromMeasureReportGroup(reportGroup);
+		// Logic for the numerator exclusion, denominator exclusion, and denominator
+		// exception has already been applied by the standard report generator and the
+		// patient has been removed from the following populations as needed, so we can
+		// keep this logic simple.
+		if (results.inInitialPopulation() && results.inDenominator() && !results.inNumerator()) {
```
jillrdoty, line 159: I think we might need to expand on this logic - given these are logically included and not define meets, we don't need inInitialPopulation() (you have to be to be in denominator).  And I think we'll also need not in numerator exclusion and denominator exception... although I'm somewhat concerned looking at this code https://github.com/DBCG/cqf-ruler/blob/master/r4/src/main/java/org/opencds/cqf/r4/evaluation/MeasureEvaluation.java#L401 that it doesn't account for denominator exclusion or numerator exclusion.

dkwasny-ibm (commented):
Everything seems OK related to my comments.  I'll leave final approval to Jill.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
csandersdev: The numerator exclusion, denominator exclusion, and denominator exception rules are definitely applied by the cqf-ruler logic. If the patient meets the exclusion criteria, they are removed from the corresponding population (e.g. numerator and numerator exclusion) and added to the exclusion population. See https://github.com/DBCG/cqf-ruler/blob/164e651c3be9a468fac6f6a5be838bd072aaba76/r4/src/main/java/org/opencds/cqf/r4/evaluation/MeasureEvaluation.java#L414 and 
https://github.com/DBCG/cqf-ruler/blob/164e651c3be9a468fac6f6a5be838bd072aaba76/r4/src/main/java/org/opencds/cqf/r4/evaluation/MeasureEvaluation.java#L207

cohort-engine/src/main/java/com/ibm/cohort/engine/translation/InJVMCqlTranslationProvider.java
```diff
@@ -25,24 +26,37 @@
...
 
+import com.ibm.cohort.engine.LibraryFormat;
+
 /**
- * Uses the CqlTranslator inprocess to convert CQL to ELM. This is
- * an alternative to using the CQL Translation Service microservice
- * which is a longer term goal and something we've tested with, but
- * are not ready to deliver.
+ * Uses the CqlTranslator inprocess to convert CQL to ELM. This is an
+ * alternative to using the CQL Translation Service microservice which is a
```
msargentibm, line 33: Is the plan to eventually use a microservice still? I feel like I remember hearing that the existing CQL translation service wasn't going to meet our needs, but I couldn't remember if we were still writing our own at some point.

cohort-engine/src/test/java/com/ibm/cohort/engine/BaseFhirTest.java
```diff
@@ -38,27 +40,32 @@
...
+
 	static int HTTP_PORT = 0;
-	
+
 	@BeforeClass
 	public static void setUpBeforeClass() {
 		// get a random local port for test use
-		try( ServerSocket socket = new ServerSocket(0) ) {;
+		try (ServerSocket socket = new ServerSocket(0)) {
+			;
```
msargentibm, line 62: Extra `;`?

msargentibm (commented):
Minor comments. I plan on doing another pass after you get a chance to present the code to the team.

cohort-engine/src/test/java/com/ibm/cohort/engine/BaseFhirTest.java
csandersdev: Yeah. This is the second one that was found in this PR. I'm not sure where they are coming from. I'm going to blame my Eclipse formatter until you can prove otherwise. :-) Will fix.

cohort-engine/src/main/java/com/ibm/cohort/engine/translation/InJVMCqlTranslationProvider.java
csandersdev: Right now, there are no plans to use the microservice, but I'm keeping the indirection in case we do end up with some other translation mechanism. Could change if people want.

cohort-engine/.settings/org.eclipse.core.resources.prefs
csandersdev: Added these to .gitignore and removed the .settings files.

cohort-engine/src/main/java/com/ibm/cohort/engine/ClasspathLibrarySourceProvider.java
csandersdev: Missed a bunch of copyrights. Added in now.

cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
```diff
@@ -71,11 +74,15 @@
  */
 public class CqlEngineWrapper {
 
+	public static final List<String> SUPPORTED_MODELS = Arrays.asList("http://hl7.org/fhir",
```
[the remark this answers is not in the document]
csandersdev: When I created the unit test for USCore, it would not function without mapping the URL in the compiled ELM (which is the USCore  URL) to the data provider. This was added to support that test case.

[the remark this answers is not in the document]
msargentibm: I get why it's there currently and think it should stay as-is for the time being. I was mostly trying to point out that I think it's the intention of the open source folks to always emit ELM that uses the FHIR url rather than the url of a specific IG. Even your test ELM for USCore contains the base FHIR URL (and the unit test seemed to work for me locally if I removed the uscore url from the map above).

I wanted to point out that we may find out that only the `http://hl7.org/fhir` url should be supported (or at least that's all that the cql tooling folks intend to support), and that it's up for the model info/cql tranlator to get that mapping right.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
jillrdoty: Thanks for the pointer... I feel like their code would have been more obvious if it handled all the scenarios the same way.  However, I feel like the first half of my comment is still valid - we don't need to check inInitialPopulation if we're checking inDenominator, and we do need to check for numeratorExclusion and denominatorException.

i.e. if a patient meets the initial population, denominator, and fails the numerator they might then meet the denominator exception criteria.  If they do, I don't think they should be considered for care gap calculation.

csandersdev: I agree that the inInitialPopulation check is redundant. I don't agree that we need additional logic for the exclusion and exception criteria. That logic was already applied.  For your example, a patient that met the denominator exception criteria would have already been removed from the denominator population by the cqf-ruler before our logic for caregaps runs. We use the output of cqf-ruler as input to our processing. This is different than if we were checking each of the criteria evaluations ourselves.

deekim:
I changed the target branch of this PR to `main` in preparation for the changing of the default branch.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
jillrdoty: You are correct about denominator exception, my thinking was wrong.  Would like to see the initial pop check removed but approving regardless.

jillrdoty (approved):

dkwasny-ibm (approved):

PR #11, opened 2020-11-03
Modified engine to use JSON.

ToddHicks:


jaigoradia:
One comment as far as env goes.  i assume before running test the FHIR server configuration json would need to be updated to point the right fhir server and tenant in the CDT environment. Approve.

jaigoradia (approved):
Approved.

deekim:
I changed the target branch of this PR to `main` in preparation for the changing of the default branch.

PR #12, opened 2020-11-06
Model info support

msargentibm:
~I originally had code where I was passing around the model info information so it would be available in the `InJVMCqlTranslationProvider`. That started to make the code pretty clunky having to modify function calls along the way, so I instead took a page out of the cql translator's book and just register a model info provider directly from `runWithArgs`.~

Edit: Code wound up in a middle ground after a couple revisions. Not quite the same place as the starting place described above.

I added a fake IG and a test CQL written against the IG. The fake IG is base FHIR + a single `FakePatient` resource I added. `FakePatient` has one custom extension called `fakeExtension`.

The test CQL references the `FakePatient` and its custom extension. I was somewhat lazy with the test, since the main thing to verify is that the test runs without hitting an exception along the way.  I'm happy to change the test or add tests if anyone  has specific suggestions.

### DIFF
--- a/cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
+++ b/cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
@@ -671,7 +671,7 @@ public class CqlEngineWrapper {
 			}
 
 			boolean isForceTranslation = arguments.sourceFormat == LibraryFormat.CQL;
-			CqlTranslationProvider<File> translationProvider = new InJVMCqlTranslationProvider(sourceProvider);
+			CqlTranslationProvider translationProvider = new InJVMCqlTranslationProvider(sourceProvider);
 			if (arguments.modelInfoFile != null && arguments.modelInfoFile.exists()) {
 				translationProvider.convertAndRegisterModelInfo(arguments.modelInfoFile);
 			}
--- a/cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
+++ b/cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
@@ -26,16 +26,11 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.zip.ZipInputStream;
 
-import javax.xml.bind.JAXB;
-
 import org.apache.commons.lang3.tuple.Pair;
-import org.cqframework.cql.cql2elm.ModelInfoLoader;
-import org.cqframework.cql.cql2elm.ModelInfoProvider;
 import org.cqframework.cql.elm.execution.ExpressionDef;
 import org.cqframework.cql.elm.execution.FunctionDef;
 import org.cqframework.cql.elm.execution.Library;
 import org.cqframework.cql.elm.execution.VersionedIdentifier;
-import org.hl7.elm_modelinfo.r1.ModelInfo;
 import org.opencds.cqf.cql.engine.data.CompositeDataProvider;
 import org.opencds.cqf.cql.engine.data.DataProvider;
 import org.opencds.cqf.cql.engine.execution.Context;
@@ -60,6 +55,7 @@ import com.beust.jcommander.Parameter;
 import com.beust.jcommander.internal.Console;
 import com.beust.jcommander.internal.DefaultConsole;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.ibm.cohort.engine.translation.BaseCqlTranslationProvider;
 import com.ibm.cohort.engine.translation.CqlTranslationProvider;
 import com.ibm.cohort.engine.translation.InJVMCqlTranslationProvider;
 
@@ -622,17 +618,6 @@ public class CqlEngineWrapper {
 		typedValue = new Quantity().withValue(new BigDecimal(parts[0])).withUnit(parts[1]);
 		return typedValue;
 	}
-	
-	protected static void registerModelInfo(File modelInfoFile) {
-		ModelInfo modelInfo = JAXB.unmarshal(modelInfoFile, ModelInfo.class);
-		// Force mapping  to FHIR 4.0.1. Consider supporting different versions in the future.
-		// Possibly add support for auto-loading model info files.
-		modelInfo.setTargetVersion("4.0.1");
-		modelInfo.setTargetUrl("http://hl7.org/fhir");
-		org.hl7.elm.r1.VersionedIdentifier modelId = (new org.hl7.elm.r1.VersionedIdentifier()).withId(modelInfo.getName()).withVersion(modelInfo.getVersion());
-		ModelInfoProvider modelProvider = () -> modelInfo;
-		ModelInfoLoader.registerModelInfoProvider(modelId, modelProvider);
-	}
 
 	/**
 	 * Simulate main method behavior in a non-static context for use in testing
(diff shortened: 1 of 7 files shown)


cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
```diff
@@ -613,6 +622,17 @@ protected static Object resolveQuantityParameter(String value) {
 		typedValue = new Quantity().withValue(new BigDecimal(parts[0])).withUnit(parts[1]);
 		return typedValue;
 	}
+	
+	protected static void registerModelInfo(File modelInfoFile) {
```
csandersdev, line 626: This would be more appropriately housed in the CqlTranslationProvider and InJVMCqlTranslationProvider. You are breaking the encapsulation here. For example, if we did use the microservice for translation, this isn't going to work.

cohort-engine/src/test/java/com/ibm/cohort/engine/CqlEngineWrapperTest.java
```diff
@@ -733,6 +735,56 @@ public void testInvalidLibraryName() throws Exception {
...
+		mockFhirResourceRetrieval(patient);
+
+		Library root = getLibrary("test", "cql/ig-test/test.cql");
+		Library helpers = getLibrary("FHIRHelpers", "cql/includes/FHIRHelpers.cql", "text/cql",
+									 "cql/includes/FHIRHelpers.xml", "application/elm+json");
+
+		RelatedArtifact related = new RelatedArtifact();
+		related.setType(RelatedArtifactType.DEPENDSON);
+		related.setResource("/Library/" + helpers.getId());
+		root.addRelatedArtifact(related);
```
csandersdev, line 755: I wouldn't block merge for this, but it would be a nice-to-have to pull this out into helper method(s) in BaseFhirTest as it something I can see being needed in multiple tests.

cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
```diff
@@ -613,6 +622,17 @@ protected static Object resolveQuantityParameter(String value) {
...
+		ModelInfo modelInfo = JAXB.unmarshal(modelInfoFile, ModelInfo.class);
+		// Force mapping  to FHIR 4.0.1. Consider supporting different versions in the future.
+		// Possibly add support for auto-loading model info files.
+		modelInfo.setTargetVersion("4.0.1");
+		modelInfo.setTargetUrl("http://hl7.org/fhir");
```
csandersdev, line 631: This feels wonky to me, but I get why you did it. It definitely simplifies the engine logic as we only need to support a data provider for the given FHIR URL and version and not every single custom IG under the sun. I was wondering if there is any chance, though, that this will conflict with the actual base FHIR modelinfo. I don't think so, but worth thinking about if you haven't already. Also, we just need to understand and agree that modelinfos for non-FHIR models are supported through this mechanism. I don't think that is a problem, but maybe something to put in the userguide.

csandersdev (changes requested):

cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
msargentibm: Best as I can tell, `targetUrl` and `targetVersion` are the mechanism that the CQF folks have in place to map back to base FHIR. Technically I am still waiting to hear back from them on what the intended behavior is. If we  need a target url and version set for the translator to  map to base  FHIR  properly, then I'd say we can either force one or just fail outright if the model info  does not already contain a target  version/url we support.

Note that `url` and `targetUrl` are two different fields in  the model info. `url` will still contain what is being mapped from (CDM url), and `targetUrl` will contain the FHIR url (what is being mapped  to). `name` and `version` should be what  is used to identify the model info itself. I don't think this should interfere with  the base FHIR modelinfo,  but  I can double check this.

cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
msargentibm: Is it ok to just move the static method, or would you like a different flow for passing along the model info and registering it?

cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
msargentibm: Screen Shot 2020-11-06 at 5 10 52 PM
Looks like once registered both base FHIR model info and the CDM model info are still available. They are keyed on name and version

cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
csandersdev: I think what you have where you pass the path for path for model info file into the provider constructor is fine. I would have put the register method implementation directly into InJVMCqlTranslationProvider and not BaseCqlTranslationProvider, though, because it only applies to that implementation. If we used the microservice, it would have to read the file and send it in the POST and would completely bypass this logic.

cohort-engine/src/main/java/com/ibm/cohort/engine/translation/InJVMCqlTranslationProvider.java
```diff
@@ -47,10 +64,17 @@ public InJVMCqlTranslationProvider(LibraryManager libraryManager, ModelManager m
 		this.modelManager = modelManager;
 		this.libraryManager = libraryManager;
 	}
-	
+
 	public InJVMCqlTranslationProvider(LibrarySourceProvider provider) {
+		this(provider, null);
+	}
+	
+	public InJVMCqlTranslationProvider(LibrarySourceProvider provider, File modelInfoFile) {
```
csandersdev, line 72: I would prefer to have a version of the translate interface that takes the modelInfoFile vs. the constructor. If we end up supporting translation in another provider, it will also need to pass in the modelinfo.

msargentibm: Any thoughts on how you would prefer the model info information to be passed along to the `TranslatingLibraryLoader` since that is what is calling `translate()`? I figured changing the `load()` signature wasn't something I should be doing since it comes from the `LibraryLoader` interface from the open source project.

csandersdev: Fair point, Thinking more about the problem, it is theoretically possible that we would need more than one custom model info file registered. Perhaps the answer is to put registerModelInfo on the CqlTranslationProvider interface. You would have to remove "static" from the implementation in InJVMCqlTranslationProvider, remove the modelInfo argument from the constructor, and then update the CohortEngine runWithArgs to call the register method on the translationProvider instance directly. If we end up supporting more than one provider that would feel pretty natural.

cohort-engine/src/main/java/com/ibm/cohort/engine/translation/CqlTranslationProvider.java
```diff
@@ -19,10 +22,26 @@
...
 	public Library translate(InputStream cql, List<Options> options, LibraryFormat targetFormat) throws Exception;
+	
+	public default void registerModelInfo(ModelInfo modelInfo) {
+		// Force mapping  to FHIR 4.0.1. Consider supporting different versions in the future.
+		// Possibly add support for auto-loading model info files.
+		modelInfo.setTargetVersion("4.0.1");
+		modelInfo.setTargetUrl("http://hl7.org/fhir");
+		org.hl7.elm.r1.VersionedIdentifier modelId = (new org.hl7.elm.r1.VersionedIdentifier()).withId(modelInfo.getName()).withVersion(modelInfo.getVersion());
+		ModelInfoProvider modelProvider = () -> modelInfo;
+		ModelInfoLoader.registerModelInfoProvider(modelId, modelProvider);
```
csandersdev: Would like this implementation to be in the InJVMTranslationProvider, not the interface.

cohort-engine/src/main/java/com/ibm/cohort/engine/translation/InJVMCqlTranslationProvider.java
```diff
@@ -120,4 +100,9 @@ public Library translate(InputStream cql, List<Options> options, LibraryFormat t
 		return result;
 	}
 
+	@Override
+	public ModelInfo convertToModelInfo(File modelInfoObject) {
+		return JAXB.unmarshal(modelInfoObject, ModelInfo.class);
+	}
+
```
csandersdev: I'm not opposed to generics, but I'm not sure this is the best use. I'd like to see several overloads that take File, InputStream, and Reader vs. using the generics for this specific.

csandersdev (changes requested):

csandersdev (approved):
Looks good now

PR #13, opened 2020-11-06
Updates to build process, new cohort-cli project, documentation refresh

csandersdev:
* Move the top-level POM into a new cohort-parent project. This simplifies integration with Eclipse and is consistent with the pattern used in many other IBM projects (e.g. IBM/FHIR).
* Add some M2E configuration to avoid POM errors in Eclipse. We ignore the flatten POM during M2E execution.
* Split the CLI part of the cohort engine wrapper into a separate project and added a similar CLI for the measure evaluator. The key driver for the split was that it separates runtime concerns for the CLI (such as binding to a specific logging implementation) from the library code. Consumers of the library won't pick up the extra runtime dependencies.
* Updated the user guide with a bunch more detail and made sure it reflects build and CLI changes made in this PR.


.github/workflows/build.yml
```diff
@@ -13,5 +13,5 @@ jobs:
         with:
           java-version: 1.8 
       - name: Build with Maven
-        run: mvn clean install
+        run: mvn clean install -f cohort-parent
```
jillrdoty, line 16: Sample comment

cohort-cli/src/main/java/com/ibm/cohort/cli/ParameterHelper.java
```diff
@@ -0,0 +1,150 @@
...
+			Matcher m = p.matcher(arg);
+			if (m.matches()) {
+				String name = m.group("name");
+				String type = m.group("type");
+				String value = m.group("value");
+
+				Object typedValue = null;
+				String[] parts = null;
+				switch (type) {
+				case "integer":
```
jillrdoty, line 44: Slight preference to see this as an enum, but I recognize that's not a common thing in the open engines codebase.

cohort-cli/src/main/java/com/ibm/cohort/cli/ParameterHelper.java
```diff
@@ -0,0 +1,150 @@
...
+				case "concept":
+					throw new UnsupportedOperationException("No support for concept type parameters");
+				case "interval":
+					parts = value.split(",");
+					String subType = parts[0];
+					String start = parts[1];
+					String end = parts[2];
+
+					switch (subType) {
+					case "integer":
```
jillrdoty, line 77: Not blocking:

It'd be nice to not see duplication here (i.e. make this a recursive call or break out that logic).  

However, I do notice that not all of the above cases are covered - if we're explicitly not allowing intervals of codes and concepts, I'd still prefer to see that handled separately and reuse the base conversation above.

cohort-cli/src/test/resources/cql/zip/Breast-Cancer-Screening.xml
```diff
@@ -0,0 +1,298 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jillrdoty, line 1: Non-blocking: I think these are all copies of the files in cohort-engine?  (I don't see them as moved or deleted there?)

It would be nice to not have duplicate resources in different projects, but I don't know of an established pattern to avoid this.

jillrdoty (approved):

cohort-cli/src/test/resources/cql/zip/Breast-Cancer-Screening.xml
csandersdev: I did some testing trying to figure out how to reuse the test assets from the cohort-engine project via Maven build hooks in the cohort-cli, but the timing was off. I could alternatively have depended directly on filesystem paths cross-project, but that is bad form and introduces a different set of problems. I settled on just duplicating the resources for now.

PR #18, opened 2020-11-17
Githubaction publish

jillrdoty:
This PR contains a couple things:
* Deploy modules to github packages.  I excluded the following [TestWrapper, build, reports, docs].  I was on the fence about TestWrapper but it was giving me an error so that decided me towards not publishing - if this is desired, I'll take another look.
* Add maven caching to the build.  My rough estimate is that this saved me ~30 seconds per build when I was building the same branch over and over.
* Add SonarCloud integration.  And updated docs for it.
* Deleted the empty cdm-modelinfo-tooling module.

*IMPORTANT NOTE*
I only configured publishing and sonar analysis on push to `main` (and releases, which we haven't really figured out officially yet).  This means branches and PRs will *not* get published or analyzed automatically.  

I can add SonarCloud to branches, but I didn't feel it added much, and slows down the build.  

Regarding publishing, I didn't feel like the very large number of versions showing up in the github packages UI would be worth it.  If others are of a different opinion, I can add the publish step where the team would like.


deekim (approved):
Nice! LGTM.

rickjstevens (approved):

PR #21, opened 2020-11-19
Sandersc config updates

csandersdev:
The QA team requested additional details in the user guide about the use
of the FHIR server configuration files. In the process of updating the
user guide with details, I realized a few things that I wanted to
improve with the configurability of the system. I made two big changes
to configuration support and added those to the docs.

1. Refactored the FhirClientFactory into a factory + builder pattern
that supports overrides of the FhirClientBuilderFactory by consumers of
the library. This allows developers an extension point to do advanced
configuration of their FHIR clients if necessary.

2. Beefed up the configurability of the default FHIR server config class
so that all of the basic HAPI client configuration features are
available. Most notably, this adds support for HAPI client logging which
has been a conspicuously missing feature during testing, but general
support has been added as well for additional headers, cookies, and HTTP bearer
token authorization as well.


cohort-engine/src/main/java/com/ibm/cohort/engine/FhirClientBuilderFactory.java
```diff
@@ -0,0 +1,57 @@
...
+	 * creates a new factory instance. The method uses the following ordered lookup
+	 * procedure to determine the FhirClientBuilderFactory implementation class to
+	 * load:
+	 * 
+	 * <ul>
+	 * <li>Use the com.ibm.cohort.FhirClientBuilderFactory</li> system property.
+	 * <li>Default FhirClientBuilderFactory instance.
+	 * </ul>
+	 * 
+	 * @return New instance of a DocumentBuilderFactory
```
jillrdoty, line 38: Should this be FhirClientBuilderFactory?

jillrdoty (approved):

cohort-engine/src/main/java/com/ibm/cohort/engine/FhirClientBuilderFactory.java
csandersdev: Yep, good catch.

PR #23, opened 2020-11-30
Thicks cli enablement

ToddHicks:


tests/src/main/python/TestDriver_MeasureCLI.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @pytest.mark.parametrize("params, targets, output, resource, measureServer",setup())
+    def test1(self, params, targets, output, resource, measureServer):
+        self.execute(params, targets, output, resource, measureServer)
+
+    # Execute submits a query and validates the return.
+    def execute(self, params, targets, output, resource, measureServer):
+        output = re.sub('@\w+', '@',output)
+        o = output.split('\n')
+        callDetails = ["java", "Xms1G", "-Djavax.net.ssl.trustStore=config/trustStore.pkcs12", "-Djavax.net.ssl.trustStorePassword=change-password", "-Djavax.net.ssl.trustStoreType=pkcs12", "-jar", jar]
```
csandersdev, line 35: Hard-coding the details to the truststore, including the password, is generally not a good idea. You should probably pass those in from the environment like you are the other options.

tests/src/main/python/TestDriver_CohortCLI.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @pytest.mark.parametrize("params, library, version, targets,output,source,expressions,measureServer",setup())
+    def test1(self, params, library, version, targets, output, source, expressions, measureServer):
+        self.execute(params, library, version, targets, output, source, expressions, measureServer)
+
+    # Execute submits a query and validates the return.
+    def execute(self, params, library, version,  targets, output, source, expressions, measureServer):
+        output = re.sub('@\w+', '@',output)
+        o = output.split('\n')
+        callDetails = ["java", "-Xms1G", "-Xmx1G", "-Djavax.net.ssl.trustStore=cohort-cli/config/trustStore.pkcs12", "-Djavax.net.ssl.trustStorePassword=change-password", "-Djavax.net.ssl.trustStoreType=pkcs12", "-jar", jar, "-f", libraries, "-l", library]
```
csandersdev, line 38: Same comment as below. Hard-coded details regarding the trust store.

csandersdev (changes requested):

tests/src/main/python/TestDriver_MeasureCLI.py
ToddHicks: Thank you!

ToddHicks: Checked in changes. Instead of in the yaml, which also get checked in. We use an environment file that already contains information that should never be checked in. Decided to add these details there.

jaigoradia (commented):
Looks good to me with changes you made that were requested by Corey.  I will let him approve the changes and you can go ahead and merge after that.

csandersdev (approved):
LGTM

PR #24, opened 2020-11-30
Removed installation guide, updated index

csandersdev:
This addresses #172.

* The installation guide was removed. It will come back at a future date
when we have information to put in it regarding Flink, REST service,
etc.
* The index contents were updated with the same basic content from the
public README.md.


jillrdoty (approved):

tabishop (approved):
Updates look good to me. Nice write up.

rickjstevens (approved):

PR #25, opened 2020-11-30
Tabishop initial rest projects integration

tabishop:
This PR has the code to include the REST projects and stubbed API endpoints. These will later be updated with the ability to call the cohorting engine, but for now, they are just stubs. 

The cohort-engine-api-web project gets built into the .war file and the cohort-engine-api project contains all the model objects that will be used by the rest api. I created separate projects in case we wanted to use the model objects within other projects, having this separated from the REST endpoint code will allow us to avoid having circular compile dependencies.

There were a lot of changes to the cohort-parent POM files. I used the template projects from the common services team for these and may need to contact that team with questions if you see things that you want changed.

Also please note the updated instructions for building while referencing the rest-service-framework projects (docs\dev-guide\getting-started.md) if you want to pull this branch and build it locally.


cohort-engine-api-web/launchers/cohort-engine-api-web-VT.launch
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?><launchConfiguration type="org.eclipse.jdt.junit.launchconfig">
```
jillrdoty, line 1: I generally prefer not to commit eclipse specific files, but I'm unfamiliar with the purpose of this one so not sure if it could be auto-generated by anyone who wishes to launch in eclipse?

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestApplication.java
```diff
@@ -0,0 +1,70 @@
...
+		// Get classes from parent
+		Set<Class<?>> classes = super.getClasses();
+
+		// Add application-specific classes here
+		// Note that any handler added here will have its static block run but not its constructor
+		// when the first URL hits the service.
+		// The handler itself is instantiated on each rest call inside that handler.
+		// Expensive initialization code should be done in the handler's static block or
+		// as a lazy initialize once during the handler's construction and stored in a
+		// static member variable.
```
jillrdoty, line 46: nitpick: looks like a mix of tabs and spaces in this file

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestConstants.java
```diff
@@ -0,0 +1,23 @@
...
+package com.ibm.cohort.engine.api.service;
+
+public class CohortEngineRestConstants {
+	public static final String SERVICE_MAJOR_VERSION = "v1";
+	public static final String SERVICE_TITLE = "IBM Cohort Engine";
+	public static final String SERVICE_DESCRIPTION = "Service to evaluate cohorts and measures";
+
+	// Provide a set of package names for swagger to search for REST APIs (comma delimited).
+	// Static assignment example
+	//public static final String SERVICE_SWAGGER_PACKAGES =  "com.ibm.watson.common.service.package1, com.ibm.watson.common.service.package2";
```
jillrdoty, line 15: Should this be removed as the variable is configured on line 19?

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestConstants.java
```diff
@@ -0,0 +1,23 @@
...
+
+	// Dynamic assignment using base package name from this class
+	private static String className = CohortEngineRestConstants.class.getCanonicalName();
+	public static final String SERVICE_SWAGGER_PACKAGES = className.substring(0, className.lastIndexOf('.'));
+
+	public static final String DARK_LAUNCHED_NEW_DATE_API_FEATURE = "new_date_api_feature";
```
jillrdoty, line 21: Curious as to the purpose for these two parameters?

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * Cohort REST service
+ *
+ *
+ */
+@Path(CohortEngineRestConstants.SERVICE_MAJOR_VERSION)
+@Api(value = "cohort")
+@SwaggerDefinition(tags={@Tag(name = "measures", description = "IBM Cohort Engine Measure Evaluation")})
+public class CohortEngineRestHandler {
+	//private static final Logger logger = LoggerFactory.getLogger(CohortEngineRestHandler.class.getName());
```
jillrdoty, line 44: I know this isn't really doing anything yet, but is there a reason to comment out the logger?

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestStatusHandler.java
```diff
@@ -0,0 +1,37 @@
...
+
+	/**
+	 * Example to show how to modify the service status text.
+	 * This method should include code that verifies the service
+	 * is running properly.
+	 */
+	@Override
+	public ServiceStatus adjustServiceStatus(ServiceStatus status) {
+
+		status.setServiceState(ServiceState.OK);
```
jillrdoty, line 31: Should the state of the service be set to the value passed in instead of OK always?

Also I'm intrigued by an endpoint that allows an external caller to update the state of the service - can you share a real-life use case for this?

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/CohortEngineRestApplicationTest.java
```diff
@@ -0,0 +1,38 @@
+package com.ibm.cohort.engine.api.service;
```
jillrdoty, line 1: Copyright needed

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/CohortEngineRestConstantsTest.java
```diff
@@ -0,0 +1,22 @@
+package com.ibm.cohort.engine.api.service;
```
jillrdoty, line 1: Copyright needed

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandlerTest.java
```diff
@@ -0,0 +1,175 @@
+/* ***************************************************************** */
```
jillrdoty, line 1: Correct copyright needed

cohort-parent/pom.xml
```diff
@@ -55,6 +58,40 @@
...
+		<whCompanyName>IBM Corp</whCompanyName>
+
+		<whBldVer>${project.version}</whBldVer>
+		<watson.buildtools.version>1.27</watson.buildtools.version>
+
+		<srcjavaDir>src/main/java</srcjavaDir>
+		<srcjavaTestDir>src/test/java</srcjavaTestDir>
+		<srcjavadocDir>${srcjavaDir}</srcjavadocDir>
+		<srcjavadocTestDir>${srcjavaTestDir}</srcjavadocTestDir>
+		<skipTests>true</skipTests>
```
jillrdoty, line 72: Curious what we're doing with setting skipTests to true?

cohort-parent/pom.xml
```diff
@@ -55,6 +58,40 @@
...
+		<outDir>${project.build.directory}</outDir>
+		<outDirName>target</outDirName>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
jillrdoty, line 75: Lines 75 and 76 are duplicates of 34 and 35?

cohort-parent/pom.xml
```diff
@@ -201,10 +243,128 @@
...
+				<version>1.2.7</version>
+				<scope>test</scope>
+			</dependency>
+			<dependency>
+				<groupId>javax.servlet</groupId>
+				<artifactId>javax.servlet-api</artifactId>
+				<version>4.0.1</version>
+				<scope>provided</scope>
+			</dependency>
+			<!-- <dependency>
```
jillrdoty, line 293: nitpick: cleanup commented dependency (I assume duplicated)

cohort-parent/pom.xml
```diff
@@ -232,13 +392,197 @@
...
+					<groupId>org.apache.maven.surefire</groupId>
+					<artifactId>maven-surefire-common</artifactId>
+					<version>3.0.0-M5</version>
+				</plugin>
+				<plugin>
+					<groupId>org.apache.maven.surefire</groupId>
+					<artifactId>surefire-junit4</artifactId>
+					<version>3.0.0-M5</version>
+				</plugin>
+<!-- 				The Failsafe Plugin is designed to run integration tests while the Surefire Plugin is designed to run unit tests.  -->
```
jillrdoty, line 496: I appreciate the plugin comments explaining their purpose.

docs/dev-guide/getting-started.md
```diff
@@ -30,7 +30,46 @@ Download [Eclipse IDE for Enterprise Java Developers](https://www.eclipse.org/do
 
 Git fork and clone the quality measure and cohort engine repository using GitHub Standard Fork & Pull Request Workflow. For instance, ``git clone git@github.com:<myuser>/Alvearie/quality-measure-and-cohort-service.git`` where ``myuser`` is your GitHub user or GitHub organization.
 
-Once cloned locally, execute the following from the root of the repository:
+The code in the quality-measure-and-cohort-service project has dependencies on code in the https://github.com/Alvearie/rest-service-framework project. In order to build, we must set up access to the snapshot packages in the rest-service-framework project. To do this, you must first create a Personal Access Token by going to https://github.com/settings/tokens and logging in with your github userid. Click "Personal access tokens" in the left side menu, then click the "Generate new token" button. Under the "Select Scopes," check read:packages, write:packages and delete:packages and click the "Generate Token" button. Copy the generated token string as it will be used in the settings.xml updates below.
```
jillrdoty, line 33: Should we be recommending write and delete packages for personal tokens?

cohort-engine-api-web/launchers/cohort-engine-api-web-VT.launch
tabishop: This file is used to launch a junit test suite from eclipse with 1 click which is very handy. It can't fully be autogenerated because it would require knowledge of the VM arguments which would be difficult to dig up for someone who isn't very familiar with the code. I would prefer to keep it in, unless it is causing a problem on other IDEs because it is something I use all the time.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestApplication.java
tabishop: Good catch, will cleanup.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestConstants.java
tabishop: It could be, but it is part of the comment illustrating of how to use include multiple packages using a comma separated list and I think contains useful information.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestConstants.java
tabishop: The rest service framework provides a handy means for having REST API calls that can be dark launched via env vars. The mechanism includes hiding them from the OpenAPI/Swagger interface when they are disabled. They would be used on the REST API definition and in the env vars to enable/disable certain API calls. These were 2 leftover examples of that from the sample application which I will cleanup.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
tabishop: Just to remove an unused variable warning in eclipse. I'm happy to uncomment it as it will be used when these REST apis are fleshed out.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestStatusHandler.java
tabishop: The actual liveness, readiness, and healthcheck endpoints are provided by the superclass ServiceStatusHandler, which is part of the rest service framework projects. The superclass endpoint methods call this method to allow the subclass to be in the implementing project and this is where the application specific logic for determining the state of the application would go. The endpoints themselves don't allow and external caller to pass in the service state. In the future we may want to add logic here to check if any configured FHIR servers, databases, or any other external services that are needed for successful operation of our service are up and can be reached.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/CohortEngineRestApplicationTest.java
tabishop: Good catch, will add.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/CohortEngineRestConstantsTest.java
tabishop: Good catch, will add.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandlerTest.java
tabishop: Good catch, will add.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
tabishop: I was wrong. Having it uncommented causes a PMD violation and fails the build <violation beginline="47" endline="47" begincolumn="37" endcolumn="42" rule="UnusedPrivateField" ruleset="Best Practices" package="com.ibm.cohort.engine.api.service" class="CohortEngineRestHandler" variable="logger" externalInfoUrl="https://pmd.github.io/pmd-6.21.0/pmd_rules_java_bestpractices.html#unusedprivatefield" priority="3">
Avoid unused private fields such as 'logger'.
</violation>

cohort-parent/pom.xml
tabishop: This was pulled over from the example project definition and I think it was an attempt to allow skipping tests on a per-project basis (ala https://maven.apache.org/surefire/maven-surefire-plugin/examples/skipping-tests.html). I don't think this is really needed for our projects at this time (we can still use the maven.test.skip property if we want in our dev envs.). I am going to remove this property and it's references until such time as we determine that it is actually needed.

cohort-parent/pom.xml
tabishop: Good catch, will cleanup.

cohort-parent/pom.xml
tabishop: Good catch, will cleanup.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
jillrdoty: Interesting, thanks

docs/dev-guide/getting-started.md
tabishop: Good point, confirmed it works with just read:packages and will remove the others

jillrdoty (approved):
Approving, but would ideally like another set of eyes on things.

docs/dev-guide/getting-started.md
```diff
@@ -30,7 +30,46 @@ Download [Eclipse IDE for Enterprise Java Developers](https://www.eclipse.org/do
 
 Git fork and clone the quality measure and cohort engine repository using GitHub Standard Fork & Pull Request Workflow. For instance, ``git clone git@github.com:<myuser>/Alvearie/quality-measure-and-cohort-service.git`` where ``myuser`` is your GitHub user or GitHub organization.
 
-Once cloned locally, execute the following from the root of the repository:
+The code in the quality-measure-and-cohort-service project has dependencies on code in the https://github.com/Alvearie/rest-service-framework project. In order to build, we must set up access to the snapshot packages in the rest-service-framework project. To do this, you must first create a Personal Access Token by going to https://github.com/settings/tokens and logging in with your github userid. Click "Personal access tokens" in the left side menu, then click the "Generate new token" button. Under the "Select Scopes," check read:packages and click the "Generate Token" button. Copy the generated token string as it will be used in the settings.xml updates below.
```
csandersdev, line 33: It appears that the github actions are working for your branch, but it isn't clear to me how without some additional work to enable these settings. Have you thought through that interaction?

cohort-engine-api-web/src/main/resources/logback.xml
```diff
@@ -0,0 +1,17 @@
...
+		class="com.ibm.watson.common.service.base.CustomAppender">
+		<encoder>
+			<pattern>%m%n</pattern>
+		</encoder>
+	</appender>
+
+	<root level="info">
+		<appender-ref ref="CUSTOM" />
+	</root>
+</configuration>
```
csandersdev, line 17: This is mostly just a comment. I jump through hoops in cohort-parent/pom.xml to remove transitive dependencies on logback from upstream OSS components because I don't believe that consumers of our library should be bound to a particular logging implementation and SLF4J likes to complain when you have more than one binding in your classpath. 

That said, I have no problem with the REST API layer being dependent on logback since it is functionally an app and nothing will link to it. I just want to make sure the logback dependency doesn't leak back into the library level code. I don't see that happening. Logback is never referenced in our pom's, so we must just be getting it transitively from the framework.

cohort-engine-api-web/src/test/resources/deploymentEnv.prop
```diff
@@ -0,0 +1,9 @@
+## setup the properties for the host under test
+## properties from this file will be picked up
+## if none are passed to the framework through
+## jvm arguments
+
+watson.service.hostname=localhost
+watson.service.port=9101
```
csandersdev, line 7: Hard-coded port? Is this specific to your local Liberty setup or will it be the same for every consumer of the code?

csandersdev (approved):
I left a few minor questions/comments, but I don't see anything that should block merge. We will probably want to rereview the API methods that we are exposing at some point, but I am fine as it is for now.

csandersdev (commented):
I'm getting a warning for "overridding managed version" in reports/pom.xml for the Maven deploy plugin. You can fix by just removing the version. It is set in cohort-parent/pom.xml now.

cohort-engine-api-web/src/main/resources/logback.xml
tabishop: The logback dependency is something provided by the rest service framework and I copied it in based on how the example application used it. It is new since the last time I used the framework and I will not claim to be an expert in it. So your comments are welcome, we should keep an eye out to see if it causes any problems once we start calling the cohort engine jar from the rest api, as I am currently packaging the cohort jar in the .war, but not yet calling it.

cohort-engine-api-web/src/test/resources/deploymentEnv.prop
tabishop: I will have to dig into this a bit more. It does get packaged with the tests jar that is intended to be used during the IBM Cloud toolchain deployment process, but unsure if it is actually used. It was pulled over from the example application.

tabishop: Talked with Tom P. and determined that this file is no longer used so I'm removing it.

docs/dev-guide/getting-started.md
tabishop: Using GitHub Actions uses a GITHUB_TOKEN which is created automatically by GitHub. I added a reference to the build.yml to make sure it can use it and authenticate.

PR #26, opened 2020-12-01
Added CLI test scenarios to test cohort cql for breast, colorectal an…

jaigoradia:
…d lung cancer screenings


ToddHicks (approved):
This looks pretty good. The naming of the scenarios is a bit odd, but we can adjust that as necessary going forward.

PR #30, opened 2020-12-08
Update powermockito to 2.0.9

msargentibm:
Updated powermockito dependency to 2.0.9. The automatic build in this repo succeeded for Java 8: https://github.com/Alvearie/quality-measure-and-cohort-service/actions/runs/408767077

I forked this repo and set up a build with Java 11 in the fork. Before the dependency update I was getting the same error Jai was hitting when building with Java 11: https://github.com/msargentibm/quality-measure-and-cohort-service/runs/1501344501?check_suite_focus=true#step:5:38631

With the dependency update, the Java 11 build succeeds: https://github.com/msargentibm/quality-measure-and-cohort-service/runs/1518725388?check_suite_focus=true

### DIFF
--- a/.github/workflows/buildJava11.yml
+++ b/.github/workflows/buildJava11.yml
@@ -1,0 +1,25 @@
+name: Java CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest 
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.11
+      - name: Cache Maven packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Build with Maven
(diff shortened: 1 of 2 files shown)


csandersdev:
Can you take a few moments and setup a second build job that runs with Java 11? The FHIR team is doing this for several different configurations and it makes a lot of sense to me. Should be fairly straight-forward I think.

csandersdev (approved):
I don't know that I ever quite noticed this before, but it looks like the build is duplicated (once each for the push and the pull-request). That is something of a separate issue from what you are trying to do, so I'm going to just approve, but would be nice to address at some point (assuming it is possible to address).

tabishop (approved):
Looks good thanks for looking into this!

msargentibm:
@csandersdev agree that we should figure out a way to prevent duplicate builds. Googled around a bit and it seems like there are suggestions from folks who have had similar issues in their projects. Would need to take time to try out some of the suggestions and verify that we still get automatic builds when we would want them.

I added this as a topic in an upcoming meeting.

jillrdoty (approved):
I like the idea of having the Java 11 build, thanks for adding.  Re Corey's comment about double builds - Kwas looked into this before and apparently it's just a thing, no easy way to prevent it.  We can chat more at sync.

PR #34, opened 2020-12-14
Update README for tests subproject

msargentibm:
Edits I made to the `tests` project README while running the tests in our CDT cluster. Edits are some formatting, bringing attention to the multiple Dockerfiles in the project, and adding a suggestion for the Env.txt file. I should note that I guessed at the purpose of the non-port-forwarding Dockerfile, so please let me know if it has a separate purpose.

I also added  a page to the internal wiki that has specific values to  use to run tests in our CDT environment on IBM Cloud: https://github.ibm.com/watson-health-cohorting/internal-wiki/wiki/Environment--Variables-for-running-the-'tests'-sub-project. I wasn't sure if we wanted to link to the wiki in our public docs/readme files or not, but can add a  link if that is something the team would think is useful.


jillrdoty (approved):

jaigoradia (commented):
Just a  minor comment.  When running tests using the docker image, I think it would be better to say run against ./tests/scenarios/{test-scenarios}.yaml instead of sampleJson.yaml as you have it.  Where {test-scenarios} is the name of a yaml file that includes the specific test scenarios to be run.

PR #35, opened 2020-12-15
De-duplicate builds for PRs.

msargentibm:
Two main changes:
* Consolidate  build files to a single file and use a build matrix.
* Add `if` statement to skip duplicate builds that would be triggered for internal PRs (only the  'push' build event  will  run, the 'pull_request' event will be skipped).

The repo settings may need to be adjusted so the correct build checks are applied to our PRs in case Github still is set up to check builds based off of the previous versions of the yaml files.


csandersdev (approved):
LGTM. Thanks for sticking with it.

jillrdoty (approved):

deekim (approved):

PR #38, opened 2021-01-11
Added new tests for testing error paths WFFHCOHORT-6 (API supports mu…

jaigoradia:
…ltiple measures)  plus updated automation framework to generate formatted results xml file

### DIFF
--- a/tests/scenarios/breastCancerTestScenarios.yaml
+++ b/tests/scenarios/breastCancerTestScenarios.yaml
@@ -17,4 +17,4 @@ scenarios:
 
 reporting:
 - module: junit-xml
-  filename: ../results/sampleTest.xml
+  filename: /bzt-configs/tests/results/sampleTest.xml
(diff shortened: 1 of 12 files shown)


tests/scenarios/breastCancerTestScenarios.yaml
```diff
@@ -17,4 +17,4 @@ scenarios:
 
 reporting:
 - module: junit-xml
-  filename: ../results/sampleTest.xml
\ No newline at end of file
+  filename: /bzt-configs/tests/results/sampleTest.xml
```
deekim, line 20: non-blocking request

I'm going to assume this works based on the example in the [README](https://github.com/Alvearie/quality-measure-and-cohort-service/blame/7961c2f5c76e45f10d58bd83f3c8e84369455095/tests/README.md#L37),
but I feel like it should at least be better documented?

deekim (approved):
lgtm, but I'd be looking to @ToddHicks since I'm not super familiar with these types of configuration files.

tests/scenarios/breastCancerTestScenarios.yaml
ToddHicks: Good point, this change was made so the result xml file would be put in the shared drive in the docker container. Jai, want to add some details in the README about some of these settings?

ToddHicks (approved):
Changes look good to me.

PR #39, opened 2021-01-12
WFFHCOHORT-199 - Add measure identifiers and resolution logic

msargentibm:
Previously you had to provide a resource id in order to evaluate a measure. This PR makes it so you can specify an identifier (with a required value and  optional system) instead of an id if needed. Out of the measures matching the provided identifier fields,  the measure with the latest semantic version will be evaluated. There is also  an option to provide a version to evaluate (regardless of which other versions of a measure exist).

There are code,  documentation and test updates. I have mixed feelings with how I altered MeasureResolutionProvider.java  and the class implementing that interface. If folks have any suggested ways to  refactor please let me know.


msargentibm:
The build is failing, but that is repo-wide at this point due to a  transitive dependency issue. I will rebase this branch once a fix is back on main.

cohort-cli/src/main/java/com/ibm/cohort/cli/input/MeasureContextProvider.java
```diff
@@ -34,6 +46,6 @@
 			parsedParameters = ParameterHelper.parseParameterArguments(parameters);
 		}
 
-		return Collections.singletonList(new MeasureContext(resourceId, parsedParameters));
+		return Collections.singletonList(new MeasureContext(resourceId, parsedParameters, null, null));
```
deekim, line 49: would prefer an overloaded constructor to be added to `MeasureContext`

```suggestion
		return Collections.singletonList(new MeasureContext(resourceId, parsedParameters));
```

deekim (approved):
lgtm, no blockers

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureVersion.java
```diff
@@ -0,0 +1,82 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.cohort.engine.measure;
+
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+public class MeasureVersion implements Comparable<MeasureVersion> {
```
csandersdev, line 13: This seems like logic that might be useful in other places. I would consider naming it SemanticVersion and maybe putting it either the parent package or maybe a util package.

csandersdev (approved):
I don't see anything blocking. I am a little curious what drove this change though. Was there something that we noticed that wanted to lookup by ID and version? It seems like the spec mostly wants resource ID or canonical URL. I haven't seen anything using identifier and version.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureVersion.java
msargentibm: Thanks for the suggestion. I went the subproject route and will bring it up with the team at our next meeting.

PR #42, opened 2021-01-18
refactor MeasureEvaluationSeed to be configurable

deekim:
resolves [WFFHCOHORT-149](https://jsw.ibm.com/browse/WFFHCOHORT-149)

`Draft PR` #5 should provide a bit more context.

### DIFF
--- a/cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
+++ b/cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
@@ -1,0 +1,179 @@
+package com.ibm.cohort.engine.measure.seed;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang3.tuple.Triple;
+import org.hl7.fhir.r4.model.Library;
+import org.hl7.fhir.r4.model.Measure;
+import org.opencds.cqf.common.evaluation.EvaluationProviderFactory;
+import org.opencds.cqf.common.helpers.DateHelper;
+import org.opencds.cqf.common.helpers.UsingHelper;
+import org.opencds.cqf.common.providers.LibraryResolutionProvider;
+import org.opencds.cqf.cql.engine.data.DataProvider;
+import org.opencds.cqf.cql.engine.debug.DebugMap;
+import org.opencds.cqf.cql.engine.execution.Context;
+import org.opencds.cqf.cql.engine.execution.LibraryLoader;
+import org.opencds.cqf.cql.engine.runtime.DateTime;
+import org.opencds.cqf.cql.engine.runtime.Interval;
+import org.opencds.cqf.cql.engine.terminology.TerminologyProvider;
+import org.opencds.cqf.r4.helpers.LibraryHelper;
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class MeasureEvaluationSeeder {
+
+	private final EvaluationProviderFactory providerFactory;
+	private final LibraryLoader libraryLoader;
+	private final LibraryResolutionProvider<Library> libraryResourceProvider;
+
+	private boolean disableExpressionCaching;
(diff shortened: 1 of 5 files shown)


jillrdoty (approved):
Without being all that familiar with the code, seems fine to me.

tabishop (approved):
Not familiar with this code area, but I don't see any obvious areas of concern.

csandersdev (approved):
I'd like to see some more unit tests that cover branches in the seeder.createContext method so that we don't take a hit to code coverage, but otherwise it LGTM.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
```diff
@@ -0,0 +1,179 @@
...
+
+
+	public IMeasureEvaluationSeed create(Measure measure, String periodStart, String periodEnd, String productLine) {
+		LibraryHelper.loadLibraries(measure, this.libraryLoader, this.libraryResourceProvider);
+
+		// resolve primary library
+		org.cqframework.cql.elm.execution.Library library =
+                LibraryHelper.resolvePrimaryLibrary(measure, libraryLoader, this.libraryResourceProvider);
+
+		List<Triple<String, String, String>> usingDefs = UsingHelper.getUsingUrlAndVersion(library.getUsings());
```
dkwasny-ibm, line 77: It might be worth adding a comment somewhere to help fresh people understand what `Triple<String, String, String>` actually means in this context.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	public MeasureEvaluationSeeder disableDebugLogging() {
+		this.debugMode = false;
+
+		return this;
+	}
+
+	public MeasureEvaluationSeeder disableExpressionCaching() {
+		this.disableExpressionCaching = true;
```
dkwasny-ibm, line 55: Super Pedantic: These booleans are a little confusing.  I would change this to be a positive boolean (like `debugMode` above) and set the value to `false`.
Swapping them to both be negative booleans (i.e. `disableDebugMode`) would work too.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
```diff
@@ -0,0 +1,179 @@
...
+
+		for (Map.Entry<Triple<String, String, String>, DataProvider> dataProviderEntry : dataProviders.entrySet()) {
+			Triple<String, String, String> usingDef = dataProviderEntry.getKey();
+			DataProvider dataProvider = dataProviderEntry.getValue();
+
+			context.registerDataProvider(usingDef.getRight(), dataProvider);
+		}
+
+		context.setParameter(
+		        null,
```
dkwasny-ibm, line 118: Super Pedantic: Is this indented a little too far?
Actually, it looks like this and the following few lines use space indentation instead of what the rest of the file uses.

dkwasny-ibm (approved):
Nothing worth blocking the PR over.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
deekim: this is actually code I had to copy from the `cql-engine` code.
So I'm not even sure I fully understand the purpose.

https://github.com/Alvearie/quality-measure-and-cohort-service/pull/5#discussion_r512787332 has the conversation on it for more context.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
deekim: this is also copied code is why that happened fwiw

PR #44, opened 2021-01-19
Wffhcohort 65 create restful artifact

tabishop:
This pull request includes a cohort-engine-distribution project which creates 2 tar.gx files, one for application binaries (ie .war file) and related files needed to run the liberty server, and another one which contains the test classes, scripts, and dependencies that are pulled from the various project files using annotations defined in the rest-service-framework dependencies (see cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java for an example of the annotations like BVT.class). The test harness can be executed via command line after you exec onto the pod in the cloud, Contact me if you would like the commands.

The .toolchain.maven.settings.xml file is used to do the maven build in the preDockerBuild.sh file so we can pull in the rest-service-framework and cql-engine dependencies which require a git token to access. The actual values are stored in key protect in the cloud and delivered via env. variables to the scripts when the toolchain runs.

It also includes a Dockerfile, helm charts, and various other files required by the toolchain to build the docker image and deploy it out onto the cloud. After this PR is merged, I can switch the toolchain to point at the main branch to fire when a PR is closed (there are other options available for triggers as well). The cloud test scripts (run-fvttests.sh, run-ivttests.sh etc.) are just stubs for now which check to make sure the liberty server is running and the status API can be called successfully. The test team will need to flesh these out with more tests in the future.

I also updated a few files I noticed the toolchain was flagging as missing copyrights. There are a lot more, but decided the rest should be another work item.


pipeline.config
```diff
@@ -0,0 +1,58 @@
+#
+# (C) Copyright IBM Corp. 2021, 2021
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+INSIGHTS:
+  ID: "9b253889-dee0-47f4-9ea0-402306239aaa"
+CI:
+  DOCKER_IMAGE_NAME: "cohorting-app"
+  UMBRELLA_REPO_PATH: "https://github.ibm.com/watson-health-cohorting/wh-cohorting-umbrella.git"
```
deekim, line 10: should this be pointing to a publicly accessible repo?

preDeploy.sh
```diff
@@ -0,0 +1,36 @@
...
+#else
+#  curl -sSL -u "watkins0@us.ibm.com:${gitApiKey}" "https://raw.github.ibm.com/whc-toolchain/njss-dev-overrides/${templateBranch}/${CLUSTER_NAME}/override.yaml" > chart/nodejs-starter/override.yaml
+#fi
+#cat chart/nodejs-starter/override.yaml
+#ls -la chart/nodejs-starter/
+
+# get this value from KP-CDT-SAMPLE Keyprotect instance
+# export REPLCOUNT="1"
+
+if echo "$PIPELINE_TOOLCHAIN_ID" | egrep "430afb18-4418-4ce9-95ea-5980bbaa7760"; then
```
deekim, line 27: Could you explain the significance of `430afb18-4418-4ce9-95ea-5980bbaa7760`?

preDeploy.sh
```diff
@@ -0,0 +1,36 @@
...
+# (C) Copyright IBM Corp. 2021, 2021
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+set -xe
+echo "pre deploy code goes here"
+
+echo "IDS_PROJECT_NAME=${IDS_PROJECT_NAME}"
+
+#templateBranch="master"
```
deekim, line 11: Can we delete some commented out code here?
This `pr` is massive, and it would go a long way in making it more consumable.

deekim (commented):
I made a pass looking through this.
I didn't see any red flags, but at the same time, this `PR` is quite large.
I don't know that I could meaningfully review this without your splitting this up into multiple `PR`s or splitting the commits.

pipeline.config
tabishop: No, we had discussed in a meeting that we wanted to point to our existing internal umbrella repo that was also being used for our FHIR server.

preDeploy.sh
tabishop: This file came from the whc-starter app which is used as a starting point for these toolchain projects. I had been advised to leave that in, although the person I reviewed it with didn't recall exactly the bug ticket, they thought it was related to a toolchain bug.

preDeploy.sh
tabishop: I opted to leave it in for now. Since the toolchain work is not 100% complete, a lot of this will need to be uncommented and modified to work with our project once we set up our helm overrides for our different deployment environments (ie CSP) and I didn't want to have to go hunt it down.

deekim: okay, I think that we should add a `todo` here with a note to give some context to the people not as familiar toolchains.

tabishop: fair point, I'll add some verbage

cohort-cli/pom.xml
```diff
@@ -1,3 +1,8 @@
+<!--  ************************************** -->
+<!--  (C) Copyright IBM Corp. 2020, 2020     -->
```
jillrdoty, line 2: I know you mentioned something about copyright cleanup so if making these all say 2021 later is an item I'll let them go.

cohort-cli/src/test/resources/cql/zip/FHIRHelpers.xml
```diff
@@ -1,4 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--  ************************************** -->
```
jillrdoty, line 2: Maybe worth talking about at standup - we didn't author these CQL files, we copied them.  Putting an IBM copyright on them feels wrong.

cohort-cli/pom.xml
tabishop: Cleaned up the ones in these files, at least

cohort-cli/src/test/resources/cql/zip/FHIRHelpers.xml
tabishop: Didn't realize they were non-IBM authored, I will remove.

Dockerfile
```diff
@@ -0,0 +1,126 @@
...
+ENV PATH="$JAVA_HOME/jre/bin:${PATH}:$ANT_HOME/bin"
+
+# Copy our startup script into the installed Liberty bin
+COPY --from=builder $COHORT_DIST_SOLUTION/solution/bin/runServer.sh $WLP_HOME/bin/
+
+# Change to root so we can do chmods to our WH user
+USER root
+
+# Grant write access to apps folder and startup script
+RUN chmod -R u+rwx,g+rwx,o+rwx $WLP_HOME && \
```
csandersdev, line 111: 777 permissions is generally a bad idea. I don't know what the standard is for docker containers and what user we get when we do something like ``kubectl exec -it ... --bash``, but in theory it would be better to grant write+execute to only the whuser account.

cohort-engine-distribution/src/main/bin/jvm.options
```diff
@@ -0,0 +1,17 @@
...
+# This file contains options that are passed directly to the JVM on startup
+# See the InfoCenter for full documentation
+
+# Uncomment the following line to enable verbose output for class loading
+#-verbose:class
+
+# Uncomment the following line to enable debug messages during VM bootstrapping
+#-Dws.ext.debug
+
+-Xmx16G
```
csandersdev, line 15: Was this intentionally 16GB? I'm not clear on how much we need, but that seems like an awful lot.

preDeploy.sh
```diff
@@ -0,0 +1,42 @@
+#
+# (C) Copyright IBM Corp. 2021, 2021
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+set -xe
+echo "pre deploy code goes here"
```
csandersdev, line 7: Nitpick. Boiler plate output is kind of ugly. Something like "Executing preDeploy.sh" would be preferable.

csandersdev (commented):
Added a few thoughts. Generally, though, LGTM.

Dockerfile
tabishop: I had copied this from a Dockerfile written by another team using a liberty server I was using as a template, but your point is well taken. I think granting access just for whuser is better. I will give it a try and see if I can get that to work.

cohort-engine-distribution/src/main/bin/jvm.options
tabishop: Good catch. No we probably don't need this much. This was the value in the template from the service -example app I used as a template for these projects. I will lower it down.

preDeploy.sh
tabishop: Happy to change it.

csandersdev (approved):
LGTM

PR #46, opened 2021-01-21
Fix error message String format

luisgarcc:
The error message that gets thrown when expected library parameters are
not provided is supposed to include the name of those missing
parameters, but the corresponding *format specifier* is missing.

This commit simply adds it.

Before:

```
Loading libraries from ZIP 'C:\Users\LuisGarcia\Downloads\CMS645v2.zip'
[main] INFO ca.uhn.fhir.context.FhirContext - Creating new FHIR context
for FHIR version [R4]
Exception in thread "main" java.lang.IllegalArgumentException: Missing
parameter values for one or more non-default library parameters: {}
```

After:

```
Loading libraries from ZIP 'C:\Users\LuisGarcia\Downloads\CMS645v2.zip'
[main] INFO ca.uhn.fhir.context.FhirContext - Creating new FHIR context
for FHIR version [R4]
Exception in thread "main" java.lang.IllegalArgumentException: Missing
parameter values for one or more non-default library parameters:
[Measurement Period]
```

### DIFF
--- a/cohort-engine/src/main/java/com/ibm/cohort/engine/LibraryUtils.java
+++ b/cohort-engine/src/main/java/com/ibm/cohort/engine/LibraryUtils.java
@@ -115,7 +115,7 @@ public class LibraryUtils {
 
 		if (!missingParameters.isEmpty()) {
 			throw new IllegalArgumentException(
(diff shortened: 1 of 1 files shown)


deekim (approved):
Thanks for the contribution!

PR #49, opened 2021-01-29
Initial draft of a measure importer tool

csandersdev:
This PR contains work in a new project - fhir-resource-tooling - that can be used to import a Measure bundle that was exported from the measure authoring UI into a FHIR server. The new project builds a shaded JAR and that JAR can be invoked simply by specifying the same JSON config file one would use to invoke the Cohort or MeasureCLI and a list of ZIP files to process. 

In order to enable the use of the FHIR configuration mechanism from the CLI projects, I refactored that code out into a new project fhir-client-config. This refactor caused me to need to touch a bunch of files for import purposes. As a side effect of the refactor, code coverage dipped below established thresholds, so I added unit tests to get us back up to a reasonable level. While trying to generate the coverage reports so that I could figure out what to add, I noticed that the mvn:site plugin was failing on the javadoc plugin. I updated the javadoc comments where necessary to get that to pass.


jaigoradia:
@csandersdev : Could you provide an example invocation of the importer tool : "The new project builds a shaded JAR and that JAR can be invoked simply by specifying the same JSON config file one would use to invoke the Cohort or MeasureCLI and a list of ZIP files to process."  Thanks.

csandersdev:
@jaigoradia I will loop back and put this information into the user guide before I deliver. I skipped that step because we aren't in final form right now and the intention was just to give Matt something he can use to work with IHE. That said, here is an example...

`java -jar fhir-resource-tooling/target/fhir-resource-tooling-0.0.1-SNAPSHOT-shaded.jar -m local-fhir-config.json /path/to/export.zip`

fhir-resource-tooling/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+		</dependency>
+		<dependency>
+			<groupId>com.github.tomakehurst</groupId>
+			<artifactId>wiremock-jre8</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-simple</artifactId>
+			<version>1.7.30</version>
```
dkwasny-ibm, line 73: Are we at the point where we should lock the `slf4j` version in `cohort-parent`?  Lots of submodules are using it these days.

cohort-engine/src/test/java/com/ibm/cohort/engine/ZipStreamLibrarySourceProviderTest.java
```diff
@@ -0,0 +1,28 @@
...
+import org.junit.Test;
+
+public class ZipStreamLibrarySourceProviderTest {
+	@Test
+	public void testLibraryFoundInZipSuccess() throws Exception {
+		try ( ZipInputStream zip = new ZipInputStream( new FileInputStream( "src/test/resources/cql/zip/breast_cancer_screening_v1_0_0_cql.zip") ) ) {
+		
+			ZipStreamLibrarySourceProvider provider = new ZipStreamLibrarySourceProvider(zip);
+			try( InputStream is = provider.getLibrarySource( new VersionedIdentifier().withId("FHIRHelpers") ) ) { 
+				assertNotNull( is );
```
dkwasny-ibm, line 24: Shut me down on this one if I'm getting too pedantic, but I'd like to see some form of content verification here.
Maybe consume the stream, hash it, and compare with an expected value, or just search the stream for `FHIRHelpers`.
Asserting non null doesn't save us from bugs in the future where the wrong content is returned for whatever reason.

dkwasny-ibm (approved):
Got some minor questions, but nothing worth blocking the PR over.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/MeasureImporter.java
```diff
@@ -0,0 +1,287 @@
...
+	 */
+	public abstract static class ArtifactMetadata {
+		public String resourceType;
+		public MetadataResource resource;
+		public String url;
+		public String version;
+		public String name;
+		public String id;
+
+		public void fromResource(MetadataResource metadata) {
```
deekim, line 61: I find this class design to be somewhat confusing, in terms of contract.

From what I can tell, `LibraryArtifact:: updateDependencies` requires `resource` to be non-`null`,
but nothing is enforcing that you call `fromResource` before `updateDependencies`.

Maybe it makes sense to make this a Factory method instead?

deekim (commented):
I agree with some of Kwas' concerns, but I wouldn't hold up a merge for any of those or my concerns.

fhir-resource-tooling/src/test/java/com/ibm/cohort/tooling/fhir/MeasureImporterTest.java
```diff
@@ -0,0 +1,185 @@
...
+					bundleEntry.setResource(resource);
+					
+					Bundle bundle = new Bundle();
+					bundle.addEntry(bundleEntry);
+					
+					mockFhirResourceRetrieval( searchUrl, bundle );
+					
+					String localUrl = "/" + resource.fhirType() + "/" + id;
+					String fullUrl = fhirConfig.getEndpoint() + localUrl;
+					System.out.println(localUrl);
```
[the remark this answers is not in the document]
csandersdev: It isn't that unusual for me to leave println's in test classes. I often find them helpful during debugging. I think that one is extraneous though. I rushed things out for Matt. Will remove.

fhir-resource-tooling/pom.xml
csandersdev: Agree. Will put this in cohort-parent/pom.xml#dependencyManagement.

cohort-engine/src/test/java/com/ibm/cohort/engine/ZipStreamLibrarySourceProviderTest.java
csandersdev: Not pedantic. This was quick and dirty to improve test coverage so that I could get Matt something usable. I'll loop back and beef it up.

fhir-resource-tooling/src/test/java/com/ibm/cohort/tooling/fhir/MeasureImporterTest.java
```diff
@@ -0,0 +1,185 @@
...
+		
+		mockFhirResourcePut("/Library/" + libraryId, "2");
+		
+		String measureId = UUID.randomUUID().toString();
+		
+		mockFhirResourcePost("/Measure", measureId, "1");
+		
+		mockFhirResourcePut("/Measure/" + measureId, "2");
+		
+		File tmpFile = new File("target/fhir-stub.json");
```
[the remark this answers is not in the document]
csandersdev: I've had issues with createTempFile in tests classes in the past. Something about the test files not getting deleted or something. It was so long ago, I don't even remember all the details. It is just institutionalized in my brain to not use it. I agree that there could be collisions. I'll poke at it and see if things have improved.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/MeasureImporter.java
csandersdev: I updated the internal classes to have private constructors.

docs/user-guide/getting-started.md
```diff
@@ -104,7 +104,7 @@ Usage: cql-engine [options]
...
...med to be R4. The ``-d`` argument to the CohortCLI is a file path pointing to a configuration file in JSON format that provides the details for connecting to the FHIR server that contains the resources that will be queried. There are two sample configuration files provided in the ``config`` folder. The ``config/remote-hapi-fhir.json`` example shows the most basic configuration which is simply the base URL of the target FHIR server. The ``config/local-ibm-fhir.json`` configuration shows a more advanced setup that uses basic authentication to connect to the FHIR server and provides a value for the IBM FHIR server tenant ID header which is used to specify which tenant in a multi-tenant environment will be used at runtime. These use the com.ibm.cohort.fhir.client.config.FhirServerConfig and com.ibm.cohort.fhir.client.config.IBMFhirServerConfig classes, respectively, and the @class property is required to indicate what type of configuration is being specified. Additional configuration is possible for developers that extend FhirServerConfig class and implement their own FhirClientBuilderFactory. A client builder factory can be configured using the com.ibm.cohort.FhirClientBuilderFactory.
```
msargentibm, line 107: In case folks are looking for the diff in the text,  I believe it's just the package names for FhirServerConfig and IBMFhirServerConfig.

msargentibm (approved):

docs/user-guide/getting-started.md
csandersdev: Correct. I waffled on whether or not I should do that because it has a lot of trickle down impact, but I figured now was much better than later.

PR #50, opened 2021-02-02
Add create methods to Parameter

msargentibm:
I was working on providing an example of passing  parameters to the `MeasureEvaluator` class for the IHE devs. We currently have the `ParameterHelper` class that can generate a parameter map of type `Map<String, Object>` to pass along to the engine during measure evaluation. There are two ways to  create the map:

1. https://github.com/Alvearie/quality-measure-and-cohort-service/blob/0501fa4e2adec384f9f82c1333ed6c786e39c470/cohort-cli/src/main/java/com/ibm/cohort/cli/ParameterHelper.java#L33
2. https://github.com/Alvearie/quality-measure-and-cohort-service/blob/0501fa4e2adec384f9f82c1333ed6c786e39c470/cohort-cli/src/main/java/com/ibm/cohort/cli/ParameterHelper.java#L73

Both options  are somewhat difficult to work with currently:

1. The string option uses a made up  format of `name:type:value` (where `value` can be further subdivided to `subtype,start,end` for  intervals).
2. The `Parameter` class does not have a straightforward way to currently create the POJO except  for using an `ObjectMapper`  to read a JSON string.

This PR adds some factory methods to the `Parameter` class to make them easier to create (which should make using the `ParameterHelper` class easier). If folks have different suggestions for making the `ParameterHelper`  class easier to interact with, I'd be open to those too. This was just something quick I wanted to get in front of people.

### DIFF
--- a/cohort-cli/src/main/java/com/ibm/cohort/cli/input/Parameter.java
+++ b/cohort-cli/src/main/java/com/ibm/cohort/cli/input/Parameter.java
@@ -7,26 +7,56 @@ package com.ibm.cohort.cli.input;
 
 import static org.apache.commons.lang.StringUtils.isEmpty;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.ibm.cohort.annotations.Generated;
 
 public class Parameter {
-	@JsonProperty("name")
+	private static final String INTERVAL_TYPE = "interval";
+
 	private String name;
-
-	@JsonProperty("type")
 	private String type;
-
-	@JsonProperty("value")
 	private String value;
-
-	@JsonProperty("subtype")
 	private String subtype;
-
-	@JsonProperty("start")
 	private String start;
-
-	@JsonProperty("end")
 	private String end;
+	
+	/**
+	 * 
+	 * @param name Name of parameter in the CQL
+	 * @param type CQL type of the parameter
+	 * @param value CQL-formatted string value of the parameter
+	 */
+	public static Parameter create(String name, String type, String value) {
+		Parameter p = new Parameter(name, type, value, null, null, null);
+		p.validate();
+		return p;
+	}
+
+	/**
+	 * 
+	 * @param name Name of parameter in the CQL
+	 * @param subtype Type of interval for the parameter
+	 * @param start String representation of the start of the interval
+	 * @param end String representation of the end of the interval
+	 */
+	public static Parameter createInterval(String name, String subtype, String start, String end) {
+		Parameter p = new Parameter(name, INTERVAL_TYPE, null, subtype, start, end);
+		p.validate();
+		return p;
+	}
+
+	@JsonCreator
+	private Parameter(@JsonProperty("name") String name, @JsonProperty("type") String type,
+					  @JsonProperty("value") String value, @JsonProperty("subtype") String subtype,
+					  @JsonProperty("start") String start, @JsonProperty("end") String end) {
+		this.name = name;
+		this.type = type;
+		this.value = value;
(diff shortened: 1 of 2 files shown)


tabishop (approved):
Looks good to me. No comments.

cohort-cli/src/main/java/com/ibm/cohort/cli/input/Parameter.java
```diff
@@ -7,26 +7,56 @@
...
+	 * @param name Name of parameter in the CQL
+	 * @param subtype Type of interval for the parameter
+	 * @param start String representation of the start of the interval
+	 * @param end String representation of the end of the interval
+	 */
+	public static Parameter createInterval(String name, String subtype, String start, String end) {
+		Parameter p = new Parameter(name, INTERVAL_TYPE, null, subtype, start, end);
+		p.validate();
+		return p;
+	}
```
csandersdev, line 47: My original design in cohort-engine for parameter handling was that applications consuming the API would pass in the strongly-typed objects from the CQL engine JARs. The string-based stuff was intended to handle CLI invocations where the only interface was a String. I think we drifted away from that with the implementation of the JSON interface. Is there a value in making the API users jump through loosely typed values to get to strongly typed parameter data?

msargentibm: > Is there a value in making the API users jump through loosely typed values to get to strongly typed parameter data?
The only advantage I can think of right now is that `ParameterHelper` at least yells at you if you pick illegal types for your parameters (and then creates strongly typed objects in a way  we expect).

If we want to steer folks right to strongly  typed objects do we want to offer any sort of code to make that  process easier for users? Or is it up to them to know the cql parameter types and make sure they are building their strongly typed objects properly?

cohort-cli/src/main/java/com/ibm/cohort/cli/input/Parameter.java
```diff
@@ -67,4 +97,33 @@ else if (isEmpty(value)) {
...
+	@Generated
+	public int hashCode() {
+		int result = name != null ? name.hashCode() : 0;
+		result = 31 * result + (type != null ? type.hashCode() : 0);
+		result = 31 * result + (value != null ? value.hashCode() : 0);
+		result = 31 * result + (subtype != null ? subtype.hashCode() : 0);
+		result = 31 * result + (start != null ? start.hashCode() : 0);
+		result = 31 * result + (end != null ? end.hashCode() : 0);
+		return result;
+	}
```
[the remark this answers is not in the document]
msargentibm: It's going to be generated by my IDE in either case, so I have no preference :) 

If we have the other style of equals and hashcode in the project already then I can switch.

cohort-cli/src/main/java/com/ibm/cohort/cli/input/Parameter.java
msargentibm: I guess I could just point at `ParameterHelper` as a reference, but steer folks to making the parameter objects on their own.

csandersdev: The biggest disadvantage I see with the strongly typed objects is that they carry implementation details about the underlying CQL engine implementation along with them and that it would be harder to swap out the engine impl for a different one at some unspecified later point. It is hard to say how likely that is.

Do we know what format the IHE application is going to have the parameter data in at the point where they are calling us? If they already have them as strings, maybe this isn't all that important to address right now. I certainly wouldn't block this PR.

msargentibm:
Going to close this for now. I got the sort of feedback I was looking for and am going to have users attempt to use the code as is before introducing changes.

PR #51, opened 2021-02-04
Wffhcohort 133 restful api return parameters

tabishop:
This PR adds 2 new REST endpoints to get the parameter information for a measure. The entry points in the code when a rest call is made are in CohortEngineRestHandler.getMeasureParametersById and CohortEngineRestHandler.getMeasureParameters

### DIFF
--- a/cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/model/MeasureParameterInfo.java
+++ b/cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/model/MeasureParameterInfo.java
@@ -1,0 +1,181 @@
+/*
+ * (C) Copyright IBM Corp. 2021, 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.cohort.engine.api.service.model;
+
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import io.swagger.annotations.ApiModelProperty;
+
+@Generated
+public class MeasureParameterInfo {
+
+	private String name = null;
+	private String use = null;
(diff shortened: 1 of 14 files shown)


csandersdev (approved):
As discussed in the team meeting, it is likely that this code will change when we confirm where measure parameters will be rendered. I am ok releasing it for now. Aside from that, I would have liked to see the test code use the same mock pattern for FHIR requests as used in other projects (via wiremock), but no reason to change it now. Approving.

cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/model/MeasureParameterInfo.java
```diff
@@ -0,0 +1,202 @@
...
+
+@Generated
+public class MeasureParameterInfo {
+
+	private String name = null;
+	private String use = null;
+	private Integer min = null;
+	private String max = null;
+	private String type = null;
+	private String documentation = new String();
```
Mary904, line 22: Our static analysis engine detects that your code is problematic. According to the [Performance (PERFORMANCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#performance-performance), [BC: Impossible downcast of toArray() result (BC_IMPOSSIBLE_DOWNCAST_OF_TOARRAY)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dm-method-invokes-inefficient-new-string-constructor-dm-string-void-ctor).

Creating a new `java.lang.String` object using the no-argument constructor wastes memory because the object so created will be functionally indistinguishable from the empty string constant `""`.  Java guarantees that identical string constants will be represented by the same `String` object.  Therefore, you should just use the empty string constant directly.

PR #52, opened 2021-02-05
Adding Javadoc

jillrdoty:
This looks like a lot of changes, but it's just cleaning up javadoc warnings.  I added a comment to the pom that was the only significant change.

Possibly contentious: I added javadoc generation to the maven jar phase, so that people would catch this building locally (and not have to run site or javadoc commands.

Probably contentious: I made javadoc warnings fail the build.  I did this first so I could catch them and clean them up, but after thinking about it decided that it made sense to keep our javadoc as clean as possible.  If there is a preference for me to remove this, will do so.

### DIFF
--- a/cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
+++ b/cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
@@ -179,11 +179,11 @@ public class CqlEngineWrapper {
 	 *                       specified <code>expressions</code> will be executed. At
 	 *                       least one value is required.
 	 * @param callback       callback function to be evaluated once per context per
-	 *                       executed define (required).
+	 *                       executed define (required).             
 	 */
 	protected void evaluateExpressionByExpression(final String libraryName, final String libraryVersion,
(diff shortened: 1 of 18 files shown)


cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEngineWrapper.java
```diff
@@ -179,11 +179,11 @@ public IGenericClient getTerminologyServerClient() {
 	 *                       specified <code>expressions</code> will be executed. At
 	 *                       least one value is required.
 	 * @param callback       callback function to be evaluated once per context per
-	 *                       executed define (required).
+	 *                       executed define (required).             
 	 */
 	protected void evaluateExpressionByExpression(final String libraryName, final String libraryVersion,
```
jillrdoty, line 184: There didn't seem to be any catchable exceptions in the class so I removed the throws.

cohort-parent/pom.xml
```diff
@@ -806,6 +806,20 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
```
jillrdoty, line 809: Only significant change

tabishop (approved):
Looks good, no comments.

deekim (approved):

PR #58, opened 2021-02-18
Added shaded jar to dockerfile

tabishop:
Small change to add the shaded cli engine jar to our dockerfile so that the test test can execute tests against it during the toolchain. I added the jar to our distribution tar.gz that gets created by the maven build, and then added it to the dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,6 +72,7 @@ ENV WLP_HOME=/opt/ibm/wlp \
     COHORT_TEST_SOLUTION=/app/cohortTestDistribution \
     JAVA_HOME=/opt/ibm/java
 ENV ALVEARIE_TEST_HOME=$ALVEARIE_HOME/cohortTestResources \
+    COHORT_ENGINE_HOME=$ALVEARIE_HOME/cohortEngine \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -84,7 +85,7 @@ USER root
 # Make dir for test resources
 # Update symlnk used by Liberty to new server.  Need root.
 RUN microdnf update -y && rm -rf /var/cache/yum && \
-    microdnf install -y --nodocs vim && \
+    microdnf install -y --nodocs vim openssl && \
```
dkwasny-ibm, line 88: Is `openssl` used by one of our downstream processes?

tabishop: That snuck in there with a cut and paste from some of the security/cert work I was doing in another branch. Most other images in Alvearie are adding that to their images. I was using it to see if I could use it to get the certs a different way. I can remove it from this PR if it bugs you, but it will probably come back in with the security code I am working on in another branch.

jaigoradia (approved):
LGTM

dkwasny-ibm (approved):

PR #59, opened 2021-02-19
WFFHCOHORT235 and WFFHCOHORT-265

csandersdev:
Added support for execute UI export bundles (235) and loading measures by canonical URL in the MeasureEvaluator (265). See individual commit messages for complete details.


cohort-engine/src/main/java/com/ibm/cohort/engine/helpers/CanonicalHelper.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * (C) Copyright IBM Corp. 2021
```
jillrdoty, line 2: nitpick: copyright template has 2 years

csandersdev: My understanding was that files should be dated with the year that they were created and then updated. For all new files, I've been using the 2021-only format. Happy to change if that isn't right. Do we have a copyright template stored somewhere as reference?

jillrdoty: I was going off https://test.cloud.ibm.com/docs/healthcare-framework?topic=healthcare-framework-3-1-prepare-source-code - it doesn't really call out the scenario of when the start and update are the same year, but I think in most other classes it's 2 dates

jillrdoty:
I personally feel like it's worth creating another package within com.ibm.cohort.engine.measure (or next to measure) for all the resolution providers... but that's just a preference thing.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/ResourceResolutionProvider.java
```diff
@@ -0,0 +1,298 @@
...
+				if (be.getResource() instanceof MetadataResource) {
+					MetadataResource metadataResource = (MetadataResource) be.getResource();
+					if (metadataResource.getId() == null) {
+						throw new IllegalArgumentException("Bundle resources must contain an ID value");
+					}
+
+					addToIndexes(metadataResource);
+				}
+			}
+		}
```
deekim, line 98: does it makes sense to:
```java
throw UnsupportedSomethingExeception();
```
here?

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/RestFhirResourceResolutionProvider.java
```diff
@@ -0,0 +1,76 @@
...
+					retVal = measure;
+					currentMax = measureVersion;
+					numberOfMaxVersionSeen = 1;
+				} else if (measureVersion.equals(currentMax)) {
+					numberOfMaxVersionSeen += 1;
+				}
+			}
+		}
+	
+		if (numberOfMaxVersionSeen != 1) {
```
deekim, line 69: just wanted to confirm that the expected behavior here is 
to return `empty` if you see `numberOfMaxVersionSeen > 1`.
(not sure if that's even possible with our `fhir` profile)

csandersdev: This logic was something that @msargentibm developed for RestFhirMeasureResolutionProvider and then I pulled it up into RestFhirResourceResolutionProvider in order to share it with the LibraryResolutionProvider. I'll give him first crack at explaining the intended semantics.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/ResourceResolutionProvider.java
csandersdev: Yes, good catch. We should throw an error if the resource type isn't handled.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/RestFhirResourceResolutionProvider.java
msargentibm: My intention was to either return a single measure or nothing.

If it's a single measure, it should be the newest semantic version.

csandersdev: Looking for at the code this morning, I think @deekim is on to something here. The resolveMeasureByIdentifier method does further validation after calling this method and will throw an Exception if no version is resolved. We should probably be doing that consistently in all the methods that now use it. I will update the code.

docs/user-guide/getting-started.md
```diff
@@ -195,8 +195,8 @@ $ java -classpath cohort-cli/target/cohort-cli-0.0.1-SNAPSHOT-shaded.jar com.ibm
         additional parameterized elements separated by comma. Multiple 
         parameters must be specified as multiple -p options
       -r, --resource
-        FHIR Resource ID for the measure resource to be evaluated. Cannot be 
-        specified if -j option is used
+        FHIR Resource ID or canonical URL for the measure resource to be evaluated.
+        Cannot be specified if -j option is used
       -t, --terminology-server
         Path to JSON configuration data for the FHIR server connection that will 
         be used to retrieve terminology.
```
msargentibm, line 202: Would you want a blurb about the `--filter` option that was added?

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/MeasureImporter.java
```diff
@@ -273,6 +224,7 @@ public static void runWithArgs(String[] args, PrintStream out) throws Exception
 
 		if (arguments.isDisplayHelp) {
 			jc.usage();
+			resultCode = 1;
```
msargentibm, line 227: Minor quesiton: Does a non-zero code make sense when a user provides the `-h` flag? Wouldn't displaying the help output be the correct behavior for the program?

msargentibm (approved):
Did my best to go through and understand all of the code. Nothing major stuck out to me.

Thanks for all of the test cases. It made looking at the code a bit easier.

PR #61, opened 2021-02-23
Load Value Sets to Fhir Server

shelleyMurphy:


fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -0,0 +1,161 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.ibm.cohort.fhir.client.config.FhirClientBuilderFactory;
+import com.ibm.cohort.fhir.client.config.FhirServerConfig;
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.parser.IParser;
+import ca.uhn.fhir.rest.api.MethodOutcome;
+import ca.uhn.fhir.rest.client.api.IGenericClient;
+
+public class ValueSetImporter {
```
dkwasny-ibm, line 33: Pedantic: I feel like this should be named `VSACValueSetImporter` just to communicate that it's purpose built to process VSAC valuesets.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -0,0 +1,161 @@
...
+				ValueSet.ConceptSetComponent component = new ValueSet.ConceptSetComponent();
+				component.setSystem(currentRow.getCell(2).getStringCellValue());
+				ValueSet.ConceptReferenceComponent concept = new ValueSet.ConceptReferenceComponent();
+				concept.setCode(currentRow.getCell(0).getStringCellValue());
+				concept.setDisplay(currentRow.getCell(1).getStringCellValue());
+				List<ValueSet.ConceptReferenceComponent> concepts = new ArrayList<>();
+				concepts.add(concept);
+				component.setConcept(concepts);
+				composeIncludes.add(component);
+			}
```
dkwasny-ibm, line 118: I think we need to have one element under `include` for each unique code system.
It looks like this code will create a unique `include` element for every row in the spreadsheet.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -0,0 +1,161 @@
...
+					default:
+						break;
+				}
+			}
+		}
+		XSSFSheet expansionSheet = wb.getSheetAt(1);
+		ValueSet.ValueSetComposeComponent compose = new ValueSet.ValueSetComposeComponent();
+		List<ValueSet.ConceptSetComponent> composeIncludes = new ArrayList<>();
+		for(Row currentRow : expansionSheet){
+			if (currentRow.getCell(0) != null && currentRow.getCell(0).getCellType().equals(CellType.NUMERIC)){
```
dkwasny-ibm, line 108: I don't believe checking for a cell type of `NUMERIC` is good enough here.
Some codesystems use alphabetical characters in their codes (e.g. ICD-10).

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -0,0 +1,161 @@
...
+		this.parser = client.getFhirContext().newJsonParser();
+	}
+
+	private void importArtifacts(List<ValueSetArtifact> valueSetArtifacts) {
+
+		for (ValueSetArtifact valueSetArtifact : valueSetArtifacts) {
+			Bundle bundle = client.search().forResource(ValueSet.class).where(ValueSet.URL.matches().value(valueSetArtifact.url))
+					.returnBundle(Bundle.class).execute();
+			if (bundle.getEntry().size() > 0) {
+				valueSetArtifact.id = bundle.getEntryFirstRep().getResource().getIdElement().getIdPart();
```
dkwasny-ibm, line 69: Was there anything else we wanted to do if the value set was already loaded prior to this run?

fhir-resource-tooling/pom.xml
```diff
@@ -71,7 +71,13 @@
...
 			<artifactId>slf4j-simple</artifactId>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>org.apache.poi</groupId>
+            <artifactId>poi-ooxml</artifactId>
+            <version>4.1.2</version>
+            <scope>compile</scope>
+        </dependency>
+    </dependencies>
```
dkwasny-ibm, line 80: Pedantic: Spaces in a tabbed file

dkwasny-ibm (changes requested):

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -0,0 +1,161 @@
...
+
+	private ValueSetArtifact createArtifact(InputStream is) throws IOException {
+		XSSFWorkbook wb = new XSSFWorkbook(is);
+		XSSFSheet mainSheet = wb.getSheetAt(0);
+		ValueSet valueSet = new ValueSet();
+		for (Row currentRow : mainSheet) {
+			if(currentRow.getCell(0) != null && currentRow.getCell(1) != null) {
+				String value = currentRow.getCell(1).getStringCellValue();
+				switch (currentRow.getCell(0).getStringCellValue()) {
+					case "Value Set Name":
```
rickjstevens, line 87: I wonder if we should be doing case insensitive compares for spreadsheet labels here? To be more tolerant on specific case for these labels when these spreadsheets are authored manually vs. downloaded from VSAC?

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -0,0 +1,161 @@
...
+					valueSetArtifact.id = outcome.getId().getIdPart();
+				}
+			}
+		}
+	}
+
+	private ValueSetArtifact createArtifact(InputStream is) throws IOException {
+		XSSFWorkbook wb = new XSSFWorkbook(is);
+		XSSFSheet mainSheet = wb.getSheetAt(0);
+		ValueSet valueSet = new ValueSet();
```
rickjstevens, line 82: Would it make sense to also verify the "tab label" for each sheet - e.g. in case user inserted another worksheet in slot 0 or 1? Or just assume we will encounter other errors if the sheet "n" doesn't contain the information we're looking for?

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
rickjstevens: I also think we should scan for row with "Code" label in cell 0 as starting point for the list of coded values for this value set - to avoid finding something in the preceding header rows that we interpret as another coded value definition.

rickjstevens (approved):

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
shelleyMurphy: I'm not sure for right now, but I assumed that once we were actually writing a loader we might have a reason to want to display the value set we're not inserting in favor of, but I'm happy to delete that whole arm. Unless there's something else we ought to be doing.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
shelleyMurphy: If we think the sheet names are more stable than their positions, I'm happy to swap to retrieving by that instead

dkwasny-ibm (approved):

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/VSACValueSetImporter.java
```diff
@@ -0,0 +1,173 @@
+package com.ibm.cohort.tooling.fhir;
```
msargentibm, line 1: Copyright block missing in new files

msargentibm (approved):
Code looked sensible  to me. Only thing I noticed was the missing copyright block in new files.

PR #65, opened 2021-02-26
New helper scripts added and modified other test files for fvt tests automation in Taurus container

jaigoradia:
FVT test automation using  Containerized Taurus image in Tool chain environment.
Following files were changed or created new to setup the automation execution of fvt tests within a containerized environment inside of the CD pipeline: 

  1. pipeline.config -- In the CD pipeline section i added 'FHIR_USER_PASSWORD' to the list of vault keys to inject out of key protect.
  2. original run-fvttests.sh script was completely modified to now call a new script called run-engine-taurus-deploy.sh  with list of tests to be executed. The run-engine-taurus-deploy.sh does most of the work to setup a Taurus pod and execute the requested tests in a loop.
  3. run-engine-taurus-deploy.yaml is used by run-engine-taurus-deploy.sh to deploy/start a new pod to execute Taurus in its container. 
  4. createTrustStore.sh is helper script that goes and creates trustStore.pkcs12 dynamically. Called from within run-engine-taurus-deploy.sh script.  
  5. Added xmlCombiner.py script under a new scripts folder under tests.  

In this solution the FHIR config jsons as well as the trustStore files are being generated with the appropriate information on the fly so no exposure of passwords in GIT. 

The cohort-cli shaded jar needed for the tests  is copied out of the running cohort-services pod into the test pod.  

The Test Pod runs the Taurus image and its only job is to sleep (upto 8 hours) while the fvt tests are executed within the container using kubectl command.  Every time the CD Tool chain runs the fvt step the old test pod from prior run gets deleted and a new one spun up for the new run.  

A single xml results file is generated at the end of a the test run to be consumed by the tool chain to upload to the DevOps Insights console for display.  The DevOps Insights Console displays how many total fvt tests  got run and the number that passed / failed.


msargentibm (commented):
Have you  considered a possible refactor of the logic in `run-engine-taurus-deploy.sh` so that pod setup is separate from test setup and execution?

It would be nice to reuse parts of that script elsewhere, but calling it as is would mean that the test pod gets torn down and redeployed (and it seemed like you were purposely trying to avoid that).

dkwasny-ibm (approved):
I wouldn't merge only on my approval, but this _seems_ to be OK from my limited viewpoint.
I only have one non-blocking thought regarding merging the test results.

tests/run-engine-taurus-deploy.sh
```diff
@@ -0,0 +1,209 @@
...
+
+#generate updated default tenant fhir config json
+cp local-ibm-fhir.json fhirconfig-default-tenant.json
+sed -i '/\"password\"/s/:.*$/: \"'"${FHIR_USER_PASS}"'\",/' fhirconfig-default-tenant.json
+sed -i '/\"tenantId\"/s/:.*$/: \"default\",/' fhirconfig-default-tenant.json
+sed -i '/\"endpoint\"/s/:.*$/: \"'"${FHIR_ENDPOINT}"'\",/' fhirconfig-default-tenant.json
+
+cd ${SCENARIOS_DIR}
+# update breastCancerTestScenarios yaml file for FHIR details, path to the cohort-cli shaded jar and the final xml file to be generated.
+sed -i '/JAR/s/:.*$/: '"${SHADED_JAR}"'/' breastCancerTestScenarios.yaml
```
csandersdev, line 105: My immediate reaction to seeing these blocks of repeating replace operations was, "there has to be a better way." It turns out that there are several better ways to do search and replace in a single execution of sed.  The one I am partial to is something like sed -i -e '...' -e '...' -e '...' where each -e is the replace operation you want to perform. The upside to an approach like this is that it will make the script run faster.  In this particular situation, I expect the tests will take up the bulk of the run time, so nothing blocking. Just FYI for future.

csandersdev (approved):
LGTM

tests/run-engine-taurus-deploy.sh
jaigoradia: Thanks for suggestion

PR #66, opened 2021-03-01
Wffhcohort 136 res tful api security

tabishop:
This PR has code in it to do a number of things:
- Setup TLS communication from the Cohort REST server and the FHIR server, this involves mounting a tls cert in a k8s secret created by a request to the comOpps team, and scripting to get certs  into the key/trust stores and updating the server config files to point at them correctly.
- Changes to the REST API to support passing a Bearer OAuth token to the FHIR Client if one is provided to the REST API. Either username/password OR OAuth can be passed in the AUTHORIZATION header (but not both at the same time as this is not supported by REST in general).
- Added a k8s service definition for the cohort service as this was a pre-req for the k8s cluster signed cert as the CN needs to match the IP address and to have a constant IP address (ie cohort-services-cohort-cohort.dev:9443)


cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/FHIRRestUtils.java
```diff
@@ -175,6 +240,11 @@ public static IGenericClient getFHIRClient(String fhirEndpoint, String userName,
 		if(authString == null || authString.trim().isEmpty()) {
 			throw new IllegalArgumentException("No data in HTTP authorization header. FHIR server credentials must be passed using HTTP Basic Authentication");
 		}
+		
+		//If OAuth authentication is being used, parse the bearer token value from the headers and return it
+		if(authString.startsWith("Bearer")) {
```
deekim, line 245:
```suggestion
		if(authString.startsWith("Bearer ")) {
```

technically, this should be `Bearer<space>`
_(pushes glasses up nose)_

deekim (approved):

cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/FHIRRestUtils.java
tabishop: I will accept your suggestion, but only because you wear glasses.

PR #68, opened 2021-03-02
WFFHCOHORT-236: In-memory cache

dkwasny-ibm:
The main goal here was to insert an in-memory cache at the `RetriveProvider` layer.

I am using [caffeine](https://github.com/ben-manes/caffeine) as my underlying cache implementation.  It's a robust re-imagining of [guava's cache](https://github.com/google/guava/wiki/CachesExplained) that has better overall performance and is better supported.  Caffeine should give us all the configuration we could ever want from an in-memory cache.


cohort-engine/pom.xml
```diff
@@ -108,6 +108,12 @@
 			<version>${project.version}</version>
 		</dependency>
 
+		<dependency>
```
jillrdoty, line 111: In the spirit of more descriptive PRs, would be interested in an overview of what this library is providing for us (I did google it but feel like having it here would be beneficial too)

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/CachingRetrieveProvider.java
```diff
@@ -0,0 +1,65 @@
...
+	private final Cache<CacheKey, Iterable<Object>> cache;
+
+	public CachingRetrieveProvider(RetrieveProvider baseProvider) {
+		this.baseProvider = baseProvider;
+		// TODO: Allow for configuration of the cache via...config file?...many params?
+		// What kind of configuration options to start with?  Max size.  Anything else?
+		this.cache = Caffeine.newBuilder()
+				.recordStats()
+				.maximumSize(1_000)
+				.build();
```
dkwasny-ibm, line 31: I'm currently thinking through ways to extract the creation of the underlying cache into a separate class that gets passed all the way into the `CachingRetrieveProvider`.
I don't like the idea of having the initialization logic living so "deep" in the evaluation class hierarchy once the persistent cache becomes a thing.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/RetrieveCacheContext.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.github.benmanes.caffeine.cache.Cache;
+
+public interface RetrieveCacheContext {
+
+	Cache<CacheKey, Iterable<Object>> newCache(String patientId);
+
+	Cache<CacheKey, Iterable<Object>> getCurrentCache();
+
+	// TODO: Is this where the cache would be persisted to a datastore???
```
msargentibm, line 17: Would you want a `RetrieveCacheContext` to know how to persist itself, or would you want a separate class that would know how to take a `RetrieveCacheContext` and persist it somewhere?

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/TransientRetrieveCacheContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class TransientRetrieveCacheContext implements RetrieveCacheContext {
+
+	private Cache<CacheKey, Iterable<Object>> currentCache;
+
+	@Override
+	public Cache<CacheKey, Iterable<Object>> newCache(String patientId) {
+		this.currentCache = Caffeine.newBuilder()
+				.recordStats()
+				.maximumSize(1_000)
```
msargentibm, line 20: Any logic around 1000, or is  this just a starting point so that you can start experimenting?

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/RetrieveCacheContext.java
```diff
@@ -0,0 +1,20 @@
...
+	void cleanupCache();
```
msargentibm, line 18: Any thoughts on what `cleanupCache` would typically mean to a class implementing this interface? Is it persisting, clearing, or deleting a cache? Maybe any of the above?

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/TransientRetrieveCacheContext.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.cohort.engine.measure.cache;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+
+public class TransientRetrieveCacheContext implements RetrieveCacheContext {
```
msargentibm, line 12: Transient in the  sense that cleanup does nothing?

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/CacheKey.java
```diff
@@ -0,0 +1,94 @@
...
+				&& Objects.equals(templateId, cacheKey.templateId)
+				&& Objects.equals(codePath, cacheKey.codePath)
+				&& Objects.equals(codes, cacheKey.codes)
+				&& Objects.equals(valueSet, cacheKey.valueSet);
+	}
+
+	@Override
+	@Generated
+	public int hashCode() {
+		return Objects.hash(context, contextPath, contextValue, dataType, templateId, codePath, codes, valueSet);
```
msargentibm, line 91: When two lists of codes have the  same values in different  orders, do you want them to hash to the same value?

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/TransientRetrieveCacheContext.java
csandersdev, line 20: Seems like a good place for externalized configuration. There are lots of cache tunings knobs that might need to be turned to effect best performance and resource utilization.

csandersdev (commented):
Left some comments, but I'm pretty pleased with how cleanly you were able to implement caching into the existing codebase. Would like to see similar modifications to the CqlEngineWrapper as well. My only major complaints are around hard-binding to a cache implementation and hard-coded tuning parameters.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/ProviderFactory.java
```diff
@@ -34,20 +36,26 @@
...
 		this.dataClient = dataClient;
 		this.terminologyClient = terminologyClient;
+		this.retrieveCacheContext = retrieveCacheContext;
 	}
 	
 	@Override
 	public DataProvider createDataProvider(String model, String version) {
 		//TODO: throw an error for an unsupported model and/or version
-		ModelResolver modelResolver = new R4FhirModelResolver();
+		ModelResolver modelResolver = MODEL_RESOLVER.get();
```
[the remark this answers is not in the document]
dkwasny-ibm: I have not tried to push any change upstream yet.  This isn't even needed for this change.  I just saw that I missed it from when I first implemented the `ThreadLocal`.  I can make a reminder to create a change upstream though.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/TransientRetrieveCacheContext.java
dkwasny-ibm: This class definitely has some work to be done regarding configuration.  I just picked a reasonable default to get off the ground.  I'll add a `TODO` to be triple sure I resolve that before merging down.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/CachingRetrieveProvider.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * (C) Copyright IBM Corp. 2020, 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.cohort.engine.measure.cache;
+
+import com.github.benmanes.caffeine.cache.Cache;
```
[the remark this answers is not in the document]
dkwasny-ibm: Hmm.  To be honest, I had not heard of JCache before this.  I'll definitely give it a look.

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
```diff
@@ -149,16 +151,24 @@ public MeasureEvaluator runWithArgs(String[] args, PrintStream out) throws Excep
...
+			RetrieveCacheContext retrieveCacheContext = new TransientRetrieveCacheContext();
 			evaluator = new MeasureEvaluator(dataServerClient, terminologyServerClient);
 			evaluator.setMeasureResolutionProvider(measureProvider);
 			evaluator.setLibraryResolutionProvider(libraryProvider);
-			
+			evaluator.setRetrieveCacheContext(retrieveCacheContext);
+
 			for( String contextId : arguments.contextIds ) {
 				out.println("Evaluating: " + contextId);
+				// TODO: Create new cache for new patient...should this be in here or within `MeasureEvaulator.evaluatePatientMeasures()`???
```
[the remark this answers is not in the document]
dkwasny-ibm: Having it at the outermost layer made it easier for me to reason about while still working through the code.  I'll move the new and clear calls into `MeasureEvaluator`.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
```diff
@@ -147,7 +153,8 @@ public MeasureReport evaluatePatientMeasure(Measure measure, String patientId, S
 		LibraryResolutionProvider<Library> libraryResolutionProvider = getLibraryResolutionProvider();
 		LibraryLoader libraryLoader = LibraryHelper.createLibraryLoader(libraryResolutionProvider);
 
-		EvaluationProviderFactory factory = new ProviderFactory(dataClient, terminologyClient);
+		EvaluationProviderFactory factory = new ProviderFactory(dataClient, terminologyClient, retrieveCacheContext);
```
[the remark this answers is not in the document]
dkwasny-ibm: I agree.  Would you like for me to try and make that refactoring here?

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/RetrieveCacheContext.java
dkwasny-ibm: My current thought process was to have the context handle both the loading from and persisting to the datastore.  We could, for example, have a `COSRetrieveCacheContext` or a `DB2RetrieveCacheContext`.  I'm not sure if that's too much responsibility for one class though.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/TransientRetrieveCacheContext.java
dkwasny-ibm: Purely a silly default to get me running.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/TransientRetrieveCacheContext.java
dkwasny-ibm: Transient in the sense that the data is never persisted outside of memory.  I could probably find a better word for that though.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/cache/CacheKey.java
dkwasny-ibm: Good question.  Since I already have to convert their `Iterable` to some concrete collection type.  Using a `Set` instead of a `List` will allow for codes in any order.  I just have to think through if there's any other implications to that.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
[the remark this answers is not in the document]
csandersdev: This is as reasonable a place as any, but I'd understand if you feel like it is out of scope.

dkwasny-ibm:
Closing the PR and will create a new one.  There's too much noise here at this point.

PR #69, opened 2021-03-03
add support for measure parameter defaults

deekim:
these changes allow for parameter defaults that are specified at the measure level
to be utilized during `cql_engine` evaluation.

---

- [WFFHCOHORT-283](https://jsw.ibm.com/browse/WFFHCOHORT-283)


deekim:
@CoreySandersOH I didn't forget about the discussion we had last week around refactor things to us a "decorator" pattern.
I kept them as `todo`s in a0014e9f83a0e20413895ebdbbf5d93e716ac190.
But since I was out sick the past couple of days, I figured that could be something I circle back to in a different _PR_.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
```diff
@@ -172,7 +193,25 @@ public MeasureReport evaluatePatientMeasure(Measure measure, String patientId, S
...
 		}
 
+		measure.getExtension().stream()
+				.filter(this::isParameterExtension)
+				.forEach(measureDefault ->
+						         seed.getContext().setParameter(
+								         null,
+								         measureDefault.getId(),
+								         toCqlObject(measureDefault.getValue())));
+
```
csandersdev, line 203: I think this is in the wrong place. This should occur _before_ the part were we set the parameters based on the parameters map provided in method invocation. The map is the override values. We want those applied after the defaults specified in the Measure resource.

deekim: ah, yep. good call.

msargentibm (approved):
Seems sensible to me

shelleyMurphy (approved):

PR #70, opened 2021-03-03
Fix extra round trip to FHIR server during Measure evaluation Library load step.

csandersdev:
Kwas noticed that after my recent PR to address measure resource resolution tasks we were getting an extra round trip to the FHIR server for each Library associated with a measure. This patch updates the caching implementation to avoid the extra hit. I also added a .gitignore file to cohort-engine-api to avoid some extraneous build artifacts from showing up as changes. See the individual commit messages for complete details.

### DIFF
--- a/cohort-engine-api/.gitignore
+++ b/cohort-engine-api/.gitignore
@@ -1,0 +1,3 @@
+messages.log
+messages*.log.zip
+LogMessages.html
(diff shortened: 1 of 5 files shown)


tabishop (approved):
looks pretty straightforward. I don't see any issues.

PR #71, opened 2021-03-04
add spike docs for measure parameter model

deekim:
_covers_ type support and gaps when converting from `FHIR` to `cql` data types.


csandersdev (approved):
LGTM

msargentibm (approved):

PR #75, opened 2021-03-09
add logic to return measure parameter defaults

deekim:


cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/FHIRRestUtils.java
```diff
@@ -168,7 +174,12 @@ public static IGenericClient getFHIRClient(String fhirEndpoint, String bearerTok
 		RestFhirMeasureResolutionProvider msp = new RestFhirMeasureResolutionProvider(measureClient);
 		Measure measure = msp.resolveMeasureById(measureId);
```
tabishop, line 175: I think we'll need to add some null checks and some exception handling here. Looks like there were some changes to the RestFhirMeasureResolutionProvider since I originally wrote this code which removed some exceptions and instead returns null if a measure doesn't exist. This will cause NPE errors.

cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/FHIRRestUtils.java
```diff
@@ -168,7 +174,12 @@ public static IGenericClient getFHIRClient(String fhirEndpoint, String bearerTok
...
 
-		return FHIRRestUtils.getLibraryParmsForMeasure(measureClient, measure);
+		List<MeasureParameterInfo> libraryParameters = FHIRRestUtils.getLibraryParmsForMeasure(measureClient, measure);
```
tabishop, line 177: Just want to confirm we are merging the parameters defined on the library objects linked to the measure AND the parameters defined in the extension on the Measure object itself. If we are including the ones from the library objects, the existing code only gets those that are linked directly to the measure and not ones that are linked to other libraries. (ie it doesn't recurse through the libraries). 

Jill's comment from  WFFHCOHORT-243 was "Discussed with cohort team today, agreed that we would move forward with putting all the parameters, including the default values, on the FHIR Measure object." That led me to believe that we weren't going to get the parms of all of the libraries linked to the measure?? But I could be mistaken. Would like clarification on what we landed on.

tabishop (changes requested):
We will need changes to handle the null checking at a minimum and I may need clarification on the expected behavior.

cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/FHIRRestUtils.java
deekim: That's a good point.
I wasn't involved in drafting the initial contract.
What should the expected response be in that case?
A `400` response code for a bad request or just an empty `MeasureParameterInfoList`?

cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/FHIRRestUtils.java
deekim: That's where I was left a little confused as well. My work in my prior ticket was specifically around _measure **defaults**_.
_e.g._

```
parameter SomeAge Integer default 10

 context Patient

 define "Over 18":
 	AgeInYears() >= SomeAge + 1
```

I wasn't sure what `IHE` intended to use some of the fields in `MeasureParameterInfo`
```java
	private String name = null;
	private String use = null;
	private Integer min = null;
	private String max = null;
	private String type = null;
	private String documentation = new String();
```
specifically `min` and `max`, from the example unit test case you had:
```java
		assertEquals("Measurement Period", parmInfo.getName());
		assertEquals("In", parmInfo.getUse());
		assertEquals("1", parmInfo.getMax());
		assertEquals("0", parmInfo.getMin().toString());
		assertEquals("Period", parmInfo.getType());
```

The usage seemed to imply general guidelines for the library
(you should have at most `1` _Measurement Period_ defined on a measure).
Whereas the work I did was to specifically around setting the **default** values in the measure.

deekim: _tl;dr_ I could see `IHE` putting parameter guidelines on libraries.
My interpretation of Jill's comment and our discussion was that our _engine_ code
would only respect defaults set on the measure during evaluation.

PR #77, opened 2021-03-10
Refactor toolchain test scripts

msargentibm:
I refactored the existing toolchain FVT test scripts to separate test pod creation from test setup and execution. This should make spinning up separate test pods for different types of tests (like performance) a bit easier without needing to copy/paste a bunch of the setup code.

`tests/create-taurus-pod.sh` will create a test pod built on top of the Taurus image and set up some of the common requirements across test types (provide access to the shaded jar, create the trust store).

`tests/run-fvttests.sh` creates a test pod using the creation script. It then performs the fvt-test-specific setup steps and overwrites relevant files in the test pod. Finally, it executes its tests and retrieves  the results.

I also refactored parts of the original scripts. Some changes include:
* Replace some duplicated sed steps with a  function.
* Change sed syntax to use different  delimiters, remove the need for escaping the `/` character, and use `"` around entire expressions.
* Remove `cd` statements and instead refer to files using the path to the file.

Edit: CD toolchain run available at https://cloud.ibm.com/devops/pipelines/tekton/9434969e-4826-4c3e-8b14-88c57f265ac7/runs/4e264633-0c7c-4f10-8489-2c34bd8142f4/cd/git-sync?env_id=ibm:yp:us-south.


dkwasny-ibm (approved):
Matt and I went through this PR prior to him posting.

tests/run-fvttests.sh
```diff
@@ -4,16 +4,90 @@
...
+populateTaurusYaml ${SCENARIOS_DIR}/coloRectalCancerTestScenarios.yaml ${COLORECTAL_XMLFILE}
+
+# lungCancerTestScenarios
+populateTaurusYaml ${SCENARIOS_DIR}/lungCancerTestScenarios.yaml ${LUNG_XMLFILE}
+
+# measureCLIExample-separate-measure-server.yaml and accompanying json file
+populateTaurusYaml ${SCENARIOS_DIR}/measureCLIExample-separate-measure-server.yaml ${MEASURECLI_XMLFILE}
+sed -i "s|\"cohort-cli/config/local-ibm-fhir.json\"|\"${DEFAULT_TENANT}\"|g" ${TEST_DIR}/src/main/resources/measureCLIExample-separate-measure-server.json
+
+# Copy scenarios into the pod
```
jaigoradia, line 52: Not sure why you need to copy scenarios into the pod since the create-taurus-pod.sh already copies everything under the tests folder down into /bzt-configs  unless i am missing something.

tests/run-fvttests.sh
```diff
@@ -4,16 +4,90 @@
...
+  echo "Found 'python3' binary"
+fi
+
+# Run the xmlCombiner.py script against all XML output copied to output/Results directory earlier.
+echo "Combining all XML output into single xml file called fvttest.xml"
+$pythonBinary ${TEST_DIR}/scripts/xmlCombiner.py ${OUTPUT_DIR}/Results/*.xml
+
+# Append the contents of the combined results xml file to the output XML file (fvttest.xml)
+echo "Generating final combined results xml file fvttest.xml for DevOps Insights Console to display"
+mv combinedResults.xml fvttest.xml
```
jaigoradia, line 93: So the combinedResults.xml file will get created in the Results folder.  Where are you in the tree structure at the point you are doing the mv operation?  certainly not in Results folder from what i can tell. May have to give full path to combinedResults.xml?  
** Also note fvttest.xml needs to be at ../$TEST_DIR level because that is where the toolchain code is looking for it.

jaigoradia:
Aside from my two comments. Everything else looks good as long as you have tested it in umbrella repo and have verified fvt tests are still working and the final DevOps Insights Console is recording them successfully.

tests/run-fvttests.sh
msargentibm: I  got rid of all  of the `cd` statements, so all of these scripts run from `../$TEST_DIR` already.

The python  script does not create `combinedResults.xml` in the `Results` folder. It creates the file in the same directory the Python script is executed from, which is still `../$TEST_DIR` in this case.

tests/run-fvttests.sh
msargentibm: The create script copies everything  under `tests/` as it appears in the repo. These copy statements replace some of the key files with their updated versions that are specific to the FVT tests.

jaigoradia (approved):
Accept your explanations to my two comments. Thanks.

PR #78, opened 2021-03-11
Added three new measure evaluation automated tests

jaigoradia:
Added 3 new tests to the fvt test bucket to cover following scenarios:
1. Execute measure by URL
2.  Execute fhir resources for a measure contained in a Measure bundle export from Authoring Tool directly via MeasureCLI interface (where resources are in zip file)
3. Execute fhir resources for a measure contained in a Measure bundle export from Authoring Tool directly via MeasureCLI interface (where resources are unzipped into folders in the filesystem)


msargentibm (approved):
Existing tests looked good to me.

Did you consider having  a test where `filters` is non-null, but `regEx` is true?

csandersdev (approved):
LGTM

PR #79, opened 2021-03-12
update measure parameter endpoint 

deekim:
### primary
_update_ `/fhir/measure/{measure_id}/parameters` endpoint
to return a `Response` object that contains parameter information from the `Measure` model (bcda4a5), 
previously expected to be on the `Library` (a0326d6)

#### measure parameter default
- pull parameter default values from `ParameterDefinition` (906efde)
→ I misunderstood the IG guidance and assumed it was directly on the `Measure` (#69)

#### exception handling
- return ~`404`~`500` when requested `measure` was not found (df4e51e)
- return exception details for `500` error codes (5628223)

---

## references

- [WFFHCOHORT-278](https://jsw.ibm.com/browse/WFFHCOHORT-278)
- [CDM Library StructureDefinition](https://pages.github.ibm.com/watson-health-fhir-server/ig-common-data-model/StructureDefinition-cdm-library.html)
- [MeasureParameter extension example](https://github.ibm.com/watson-health-fhir-server/ig-common-data-model/blob/4750f53d14e1bdec7902fe4942023b806a732777/data/resources/quality-measure/QualityMeasure_Instance.fsh#L88-L91)

```
* extension[measureParameter][1].valueParameterDefinition.name = #complianceDisplayName
* extension[measureParameter][1].valueParameterDefinition.use = #out 
* extension[measureParameter][1].valueParameterDefinition.type = #string
* extension[measureParameter][1].valueParameterDefinition.extension[defaultValue].valueString = "Pharyngitis Testing"
```

---

## requested changes
- [x] handle `null` on `getParametersForMeasureIdentifier`


cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
```diff
@@ -35,7 +37,8 @@
  * server.
  */
 public class MeasureEvaluator {
-	protected static final String PARAMETER_EXTENSION_URL = "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-parameter";
+	protected static final String MEASURE_PARAMETER_URL = "http://ibm.com/fhir/cdm/StructureDefinition/measure-parameter";
+	protected static final String PARAMETER_DEFAULT_URL = "http://ibm.com/fhir/cdm/StructureDefinition/default-value";
```
deekim, line 41: Ideally, I would move these to `CDMConstants.java` once #74 is merged back

shelleyMurphy (approved):
Kept getting distracted, but this looks sensible to me.

cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/FHIRRestUtils.java
```diff
@@ -153,7 +153,7 @@ public static IGenericClient getFHIRClient(String fhirEndpoint, String bearerTok
 		RestFhirMeasureResolutionProvider msp = new RestFhirMeasureResolutionProvider(measureClient);
 		Measure measure = msp.resolveMeasureByIdentifier(identifier, measureVersion);
```
tabishop, line 155: Think we need a null check here on the measure as well, no?

tabishop (changes requested):
Think we are missing a null check/exception throw.

cohort-engine-api/src/main/java/com/ibm/cohort/engine/api/service/FHIRRestUtils.java
deekim: Yep, good call.
I forgot there were two codepaths for getting measures.

tabishop (approved):
looks good

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
```diff
@@ -205,13 +208,27 @@ public MeasureReport evaluatePatientMeasure(Measure measure, String patientId, S
 		return evaluation.evaluatePatientMeasure(measure, seed.getContext(), patientId);
 	}
 
+	private void setDefaultValue(Context context, ParameterDefinition parameterDefinition) {
+		parameterDefinition.getExtension().stream()
+				.filter(MeasureEvaluator::isDefaultValue)
+				.forEach(defaultValue ->
+						         context.setParameter(
+						         		null,
```
msargentibm, line 216: I know this is just refactored code as part of this PR, but I wanted to confirm that `null` here makes sure that the parameter is global/not tied to a specific library.

msargentibm (approved):
Only comment is a question for one line that was part of the previous PR for parameters.

jaigoradia (approved):
Approved.

dkwasny-ibm (approved):

PR #80, opened 2021-03-15
Wffhcohort 289 cloud readiness helm configuration

tabishop:
This work item contains 3 different items:

- Inclusion of a ENABLE_DARK_FEATURES property that is retrieved from helm override files in a configuration repo that has been set up. This will allow us to enable/disable certain REST endpoints in different environments by changing the value in the helm chart

- Removal of the rest-service-framework test jar from the build and docker image. We decided to use the Taurus framework for automated REST api testing so we didn't need to use this framework. The test was including a problematic jar from a code scan perspective so needed to be removed.

- Ability to set the FHIR_CLUSTER_NAMESPACE to be different than the namespace our cohort service is deploying into. Previously, it assumed they were the same which was problematic when we wanted to deploy another cohort service instance in a different namespace (say qa namespace for example), but still use the same FHIR server which was in a namespace different from the cohort deployment (ie dev namespace). Now we have a keyProtect key which sets the FHIR_CLUSTER_NAMESPACE. If we want to have multiple FHIR deployments in the future, we can just remove the keyProtect key and it will go back to the old behavior of defaulting to looking for FHIR in the same namespace as the deployment. In 3.3.0 of the toolchains, we will be able to set a toolchain property to override the property on a deployment by deployment basis for even more flexibility.


chart/cohort/values.yaml
```diff
@@ -31,4 +31,8 @@ service:
     - 9080
     - 9443
 
-internalTlsCertSecretName: cohort-services-tls-secret
\ No newline at end of file
+internalTlsCertSecretName: cohort-services-tls-secret
+# enableDarkFeatures is used to enable or disable REST endpoints depending on the env. we are deploying to
+# there are some REST endpoints we only want enabled in a development environment. The value ends up
+# in the jvm.options file the liberty server
+enableDarkFeatures: REPLACEDBYOVERRIDE
```
msargentibm, line 38: Would there be any value to setting this feature to `false`/`all disabled` by default in  the chart? Then dark features would have to  be turned on  explicitly in an override.

jillrdoty:
Can't really speak to the helm changes but excited to see more red than green :)

pipeline.config
```diff
@@ -14,14 +14,20 @@ CI:
   REPORT_PATH: "lcov.info"
   BUILDCACHE: "false"
   REPLACEARGS: "true"
-  VAULT_KEYS_TO_INJECT: "IBM_GITHUB_PUBLIC_USERNAME,IBM_GITHUB_PUBLIC_TOKEN"
+  # FHIR_CLUSTER_NAMESPACE is set in keyProtect to override the default behavior that requires the fhir server
```
msargentibm, line 17: Something about the name`FHIR_CLUSTER_NAMESPACE` isn't  meshing well with my brain. Would `FHIR_SERVER_NAMESPACE` be an alternative here?

msargentibm (approved):
If anyone else is curious, this seems to be the override repo: https://github.ibm.com/watson-health-cohorting/wh-cohorting-config

Changes look fine overall. I only had a  couple non-blocking comments.

chart/cohort/values.yaml
tabishop: I would be fine doing that. Good suggestion.

pipeline.config
tabishop: The reason I chose FHIR_CLUSTER_NAMESPACE is that the default env var provided by the toolchain for the base namespace we are deploying to is CLUSTER_NAMESPACE, so I added FHIR to it to maintain that continuity. To me it seemed to mean the namespace in the cluster, or the fhir namespace within the cluster.

msargentibm: Works for me.

I think my brain was reading `[FHIR Cluster] namespace` (as if an entirely separate cluster existed for FHIR and this was a special namespace there). I should be able to retrain my brain to see it as  `FHIR [cluster namespace]`.  😄

PR #83, opened 2021-03-18
WFFHCOHORT-236: Implementing the retrieve caching layer

dkwasny-ibm:
Apologies for the mondo PR...

This is the first iteration of the caching layer intended to limit the number of retrieve calls that make it to the FHIR server.
The implementation provided here is in-memory only, but I hope the API will easily allow us to implement per-patient persistence.

The user is expected to process all evaluations for a patient before moving on to another patient to get the best performance.
The caching layer utilizes [JCache](https://jcp.org/aboutJava/communityprocess/final/jsr107/index.html) to abstract away the actual cache implementation.
Theoretically, users will be able to provide whatever cache implementation they want by including the respective jars in their classpath.
I use [Caffeine](https://github.com/ben-manes/caffeine) as a simple in-memory implementation of JCache for both testing and integration into our CLI and Flink tools.

I performed a large refactoring of `MeasureEvaluator` and `MeasureEvaluationSeeder` to both allow for easy integration of the cache layer while also hiding implementation details from the user if desired.
The user can still manually create all the pieces necessary for measure evaluation (see `MeasureCLI`).

I updated the example flink job to better take advantage of the new caching layer by changing the input tuples to be a patient with multiple measures instead of patient with a single measure.

Oh, and don't forget to tell Github to ignore whitespace changes for a better view of what changes for some of these files.


cohort-engine-flink-example/README.md
```diff
@@ -50,8 +51,13 @@ The Flink job will run indefinitely until manually canceled.
...
-* `--rebalanceInput` (optional): Perform a `rebalance` operation after reading the input records from Kafka.
+* `--jobName <ARG>` (optional): The name of the submitted Flink Job (Default: `cohort-engine`).
+* `--printOutputToConsole` (optional): Log the MeasureReports to standard out (Default: `false`).
+* `--rebalanceInput` (optional): Perform a `rebalance` operation after reading the input records from Kafka (Default: `false`).
     * Useful if you want to scale your compute tasks separately from your Kafka topic partitions.
-* `--readFromStart` (optional): Start streaming input records from the beginning of the Kafka topic.
\ No newline at end of file
+* `--readFromStart` (optional): Start streaming input records from the beginning of the Kafka topic (Default: `false`).
+
+* `--enableCache` (optional): Enable the retrieve cache (Default: `false`).
```
jaigoradia, line 60: @dkwasny-ibm : The names of the cache options you have in README.md here should match what they are actually defined as in the code. 
"--enable-cache"               **Not --enableCache**
"--max-cache-size"            **Not --cacheMaxSize**
"--cache-expire-on-write"  **Not --cacheExpireOnWrite**
"--enable-cache-statistics" **Not --cacheEnableStatistics.**

dkwasny-ibm: This readme is for the example flink job, which [does](https://github.com/Alvearie/quality-measure-and-cohort-service/pull/83/files#diff-67ea38f40bd2516139457e385488679a7cafce70583099eb1f90d68e7af732a7R75) use those parameter names.  I don't think there is a readme for the `MeasureCLI` program.

dkwasny-ibm: I did rename `cacheEnableStatistics` to `enableCacheStatistics` to better match the MeasureCLI parameter regardless.

jaigoradia: Sorry didn't read this close enough. I was wondering if a similar section needs to be in the User Guide. https://alvearie.github.io/quality-measure-and-cohort-service/#/user-guide/getting-started?id=fhir-server-configuration

dkwasny-ibm: Good point.  I forgot about those docs.  I will make some updates now.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/R4DataProviderFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * <p>An internal class intended to ease the burden of creating the URI to {@link DataProvider} mapping
+ * required by the {@link MeasureEvaluator}.
+ *
+ * <p>This should only be used if the {@link R4MeasureEvaluatorBuilder} is insufficient for your usecase.
+ */
+public class R4DataProviderFactory {
+
+	protected static final String FHIR_R4_URL = "http://hl7.org/fhir";
```
deekim, line 32:
```suggestion
	protected static final String FHIR_R4_URL = "http://hl7.org/fhir/R4";
```
?

not sure that it would matter in the end though if you always access it via the _constant_

deekim (approved):

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/R4DataProviderFactory.java
dkwasny-ibm: It's important that the URL matches what the cqf-ruler code resolves the `using FHIR` statement to.
https://github.com/DBCG/cqf-ruler/blob/20f09e75ce524d971042e5369afcd138a9dfd079/common/src/main/java/org/opencds/cqf/common/helpers/UsingHelper.java#L19
It's either doing an assumed hardcoding match like this, or allowing ALL using statements to resolve to the provided DataProvider (regardless of whether it's actually using FHIR or not).
I guess I could also go back towards having the DataProvider be created during evaluation, but I would still need to magically know to create an R4 DataProvider when I see that exact URL.

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
```diff
@@ -79,6 +90,18 @@
 				"--define-return-option" }, description = "Specify define return option for evaluated define statements on measure report. Defaults to NONE.  Valid options are (ALL|BOOLEAN|NONE).")
 		private DefineReturnOptions defineReturnOption = DefineReturnOptions.NONE;
 
+		@Parameter(names = { "--enable-retrieve-cache" }, description = "Enable the use of the retrieve cache.")
+		private boolean enableRetrieveCache = false;
```
csandersdev, line 94: Do we really want the cache to be disabled by default? Seems like something most people would want by default.

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
```diff
@@ -79,6 +90,18 @@
...
+
+		@Parameter(names = { "--max-retrieve-cache-size" }, description = "The maximum size the retrieve cache can grow before evictions begin.")
+		private int maxRetrieveCacheSize = 1_000;
+
+		@Parameter(names = { "--retrieve-cache-expire-on-write" }, description = "The amount of time after last write (in seconds) before a retrieve cache entry is evicted.")
+		private int retrieveCacheExpireOnWrite = 300;
+
+		@Parameter(names = { "--enable-retrieve-cache-statistics" }, description = "Enable retrieve cache statistic recording via JMX.")
+		private boolean enableRetrieveCacheStatistics = false;
```
csandersdev, line 103: Rather than have options to our program for all the various possible configuration settings of the cache, did you consider making these configurable via a configuration file? For example, something like [this reference conf file](https://github.com/ben-manes/caffeine/blob/master/jcache/src/main/resources/reference.conf.)

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
dkwasny-ibm: I can enable it by default.

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
dkwasny-ibm: That configuration file is specifically for Caffeine.
It's meant to be put on the classpath and automatically read in by Typesafe's config library instead of passed in as a parameter.
That makes it hard for users to use their own config file with our CLI program (not sure if they should regardless).
They should be fine to use their own configs within their own distributions though.

- Are you looking for me to leverage the actual configfile implementation for Caffeine, or just create my own json format for configuring Caffeine myself?
- Do you want the configuration file to be implementation agnostic?  Identifying what properties make sense in an agnostic configuration file (and implementing the logic around it) is big enough to be another ticket.

I initially hit concurrency issues when I was using the typesafe config file to create new caches, but didn't think to use it to set the default values.  I can try that if desired.

dkwasny-ibm: Actually, it looks like users can override the config file with something like `-Dconfig.file=path/to/config-file` based on the typesafe docs.
I could try to provide a config file with some altered default settings.
I'm guessing we'll still want `--enable-cache` to remain though correct?

csandersdev: My general thought here is that cache configuration is more complex than these three properties. These might be enough for a basic setup, but I was imagining a world where we would want to enable more complex cache configurations without having to add flags to our CLI for each specific setting. I know that most cache providers provide a configuration file format that can be used to externalize configuration. The above referenced file is the defaults file for caffeine and they seem to expect that an application would have an application.conf file. That is the one I would expect us to point to in some way.

msargentibm (approved):
Approving assuming the comments on MeasureCLI will work out in one way or another.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureEvaluator.java
```diff
@@ -8,87 +8,44 @@
...
-	public void setLibraryResolutionProvider(LibraryResolutionProvider<Library> provider) {
-		this.libraryProvider = provider;
-	}
-	
-	public LibraryResolutionProvider<Library> getLibraryResolutionProvider() {
-		if( this.libraryProvider == null ) {
-			this.libraryProvider = new RestFhirLibraryResolutionProvider(measureClient);
-		}
-		return this.libraryProvider;
+			MeasureResolutionProvider<Measure> measureProvider,
```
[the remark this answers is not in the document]
csandersdev: Last time I talked with Luis, he was pinning his code to a particular dated snapshot. This won't directly impact his work until he chooses to bring on the new code. That said, I agree with the sentiment. We should make sure that IHE at least knows that this new snapshot introduces API changes and have a good example ready for them. @dkwasny-ibm is the updated Flink driver how you would expect a new consumer to leverage the API changes?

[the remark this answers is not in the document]
dkwasny-ibm: I think after [this commit](https://github.com/Alvearie/quality-measure-and-cohort-service/pull/83/commits/ab883da4ebc8b19cb62507788bfcc2152f65cd39), I'm comfortable saying the flink driver shows the simplest way to leverage the new API.
This is assuming the customer knows how to create a `FhirServerConfig` object.

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
dkwasny-ibm: I think I was able to (somewhat) cleanly utilize the Typesafe/Lightbend config library after [this commit](https://github.com/Alvearie/quality-measure-and-cohort-service/pull/83/commits/933c307e989a51f39d2b1cd0a378e3d348f65568).
I had to do some work in the uberjar configuration to ensure all values in all `reference.conf` files remain intact, but everything else seemed obvious.
Let me know if this addresses your concerns.

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
dkwasny-ibm: I changed the options to enable the cache by default via [this commit](https://github.com/Alvearie/quality-measure-and-cohort-service/pull/83/commits/1817b5bbc4f9921ea7b4580dd29a84ed17ac001c).

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
csandersdev: Looks ok to me.

msargentibm (approved):
Approve again after last set of commits.

PR #85, opened 2021-03-19
Thicks rest api

ToddHicks:


tests/src/main/python/rest/tests/MeasureParametersAPITests.py
```diff
@@ -0,0 +1,31 @@
...
+import swagger_client
+from swagger_client.api.fhir_measures_api import FHIRMeasuresApi
+
+class MeasureParametersAPITest(TestClass):
+
+    def test_simpleErrorCase(self):
+        fhirMeasuresApi = FHIRMeasuresApi(swagger_client.ApiClient(self.configuration))
+        result = None
+        try:
+            resp = fhirMeasuresApi.get_measure_parameters('2021-03-15','https://fhir-internal.dev:9443/fhir-server/api/v4','default','badMeasureID')
```
csandersdev, line 16: I'd like to see the FHIR server endpoint come from some type of external property instead of being hard-coded.

tests/src/main/python/rest/tests/MeasureParametersAPITests.py
```diff
@@ -0,0 +1,31 @@
...
+            result = resp.read()
+        except Exception as e: 
+            print(str(e))
+            result = str(e)
+        assert '500' in result, 'Should contain 500 error.'
+
+    def test_simpleCase(self):
+        fhirMeasuresApi = FHIRMeasuresApi(swagger_client.ApiClient(self.configuration))
+
+        resp = fhirMeasuresApi.get_measure_parameters('2021-03-15','https://fhir-internal.dev:9443/fhir-server/api/v4','default','999',_preload_content = False)
```
csandersdev, line 26: Same as above. Endpoint is hard coded.

csandersdev (commented):
It is a bit difficult to see the custom code vs. the generated code. Would be good in the future to separate the updates to the generated code from the custom code in separate commits. For what I did see in custom code, I commented on some hard-coding that we probably want to make configurable.

tests/src/main/python/rest/tests/MeasureParametersAPITests.py
ToddHicks: Good point, even the tenant could be established somewhere are reused instead of hard coded. Wondering what other common values we may want to consider.

tests/src/main/python/rest/tests/utilities/TestClass.py
```diff
@@ -0,0 +1,38 @@
...
+import swagger_client
+import unittest
+import json
+
+
+class TestClass(unittest.TestCase):
+    #Connection Details
+    configuration = swagger_client.Configuration()
+    #Check if we should read from a json file or environment variable.
+    if os.environ['FHIR_JSON']=='False':
```
msargentibm, line 15: An alternate way to accomplish this check would be to see if `FHIR_JSON` has a value with `os.getenv()`. If it does, then treat is as a file and try to load it. Otherwise, use `FHIRUSER` and `FHIRPASS`.

My gut reaction is that the check as written is sort of special behavior that hides how to properly use `FHIRUSER` and `FHIRPASS` (since `FHIR_JSON` also needs to be set to exactly `False` and the only way to know this is to dig through this code).

msargentibm (approved):
Had one thought in TestClass, but ultimately nothing major.

tests/src/main/python/rest/tests/utilities/TestClass.py
```diff
@@ -0,0 +1,38 @@
...
+            configuration.password = data['password']
+            fhir_endpoint = data['endpoint']
+            fhir_tenant = data['tenantId']
+
+    configuration.host = os.environ['HOST']
+    
+    if os.environ['VERIFY_SSL'] == 'False':
+        configuration.verify_ssl = False
+    else:
+        configuraiton.verify_ssl = True
```
deekim, line 36:
```suggestion
        configuration.verify_ssl = True
```

tests/src/main/python/rest/tests/utilities/TestClass.py
```diff
@@ -0,0 +1,38 @@
...
+    #Check if we should read from a json file or environment variable.
+    if os.environ['FHIR_JSON']=='False':
+        #Load from Environment.
+        configuration.username = os.environ['FHIRUSER']
+        configuration.password = os.environ['FHIRPASS']
+        fhir_endpoint = 'https://fhir-internal.dev:9443/fhir-server/api/v4'
+        fhir_tenant = 'Default'
+    else:
+        #Load from JSON file and set.
+        #config/local-ibm-fhir.json <- whats the real one? See create_pod shell script.
```
deekim, line 23: I'm not sure I understand this comment.
Could you elaborate?

tests/src/main/python/rest/tests/extract/.travis.yml
```diff
@@ -0,0 +1,13 @@
+# ref: https://docs.travis-ci.com/user/languages/python
```
deekim, line 1: I don't think we actually use travis?
Is there a away to turn off the autogeneration of this file to avoid confusion?

deekim (changes requested):
I'm not sure I agree with our committing autogenerated code like this.
I feel that it would lead to unnecessary diffs when review PRs.
(happy to have it be the case, but I think that you might want to bring this up as a discussion topic to solicit team input)

In addition, could you please add to the PR description the value in having the autogenerated code in version control?
Because you are using `python` to run the tests, you wouldn't need to have the rest client code at "compile time". I think a simple alternative would be to have the code generated when you executed the tests (although I'm not familiar with best practices here)?

tests/src/main/python/rest/README.md
```diff
@@ -0,0 +1,9 @@
+For the rest tests, we use the swagger-codegen. As seen here. https://github.com/swagger-api/swagger-codegen#
+
+When new API's are added to swagger perform the following.
+1. Delete (or archive) the extract folder under /python/rest/tests
+2. Start swagger on localhost.'kubectl -n dev port-forward service/cohort-services-cohort-cohort 9443:9443'
+3. Pull the swagger JSON. 'curl -k https://localhost:9443/services/cohort/api/swagger/swagger.json > swagger.json'
+4. Run generator 'swagger-codegen generate -i swagger.json -l python -o ***/python/rest/tests/extract/' Use correct path.
+5. Note, if old api's have been updated this may cause some tests to fail.
+Never change any file in the "extract" folder. This has been designed that the extract folder is fully generated.
```
deekim, line 9: If this is truly the case I feel that this should be scripted to some degree?
And ideally run as part of a GitHub action or toolchain?

tests/src/main/python/rest/tests/utilities/TestClass.py
ToddHicks: Maybe we can meet and chat. For whatever reason I couldn't get that to work when I was trying it! Probably something really silly I was doing wrong but if you think you know that call I will definitely use it, you are right, this was a bit of a hack... though I really don't expect it to be set to 'False' except when run locally.

msargentibm:
```
import os

if os.getenv("test"):
  print("HI")
```

Running the above with `python3 test-script.py` doesn't print "HI" for me locally.  Running  `test=123 python3 test-script.py` prints "HI".

Maybe something was setting the FHIR_JSON property?

tests/src/main/python/rest/tests/utilities/TestClass.py
ToddHicks: Jai has better understanding of how we will actually load the json file and where the real one is at. So some additional massaging of this code is expected. My work was to get things started and Jai will pick it up and run with it.

PR #86, opened 2021-03-22
Add performance regression tests

msargentibm:
This PR  adds a  handful of performance scenarios that will be kicked off each time the CD pipeline is run.

The hook into the toolchain is handled in the `tests/run-perfreg.sh` file. Part of the script involves creating a test pod for performance regression tests, updating `tests/scenarios/performance/performanceRegression.yaml`, and then executing that scenario with Taurus on the test pod.

I made use of sequential executions inside of Taurus so that I did not have to juggle multiple yaml files for the performance tests. Taurus will run a single executor at a time until all of the tests are finished. I also used the pass/fail module for Taurus to set a threshold for each test. If the test takes longer than the criteria in the yaml file then the test will be marked as a failure.

Here is an example report from devops insights with test labels included:
Screen Shot 2021-03-23 at 9 17 08 AM

Edit: Removed outstanding text about test names. Kwas helped dig around and we found that Labels will appear as part of the test name in the devops insights reports (rather than overwriting the entire test name). I'll update the screenshot once I get an example run with labels included for all test cases.

Edit 2: Having issues when including labels. Tests no longer seem to fail as expected if a label is included for a test case (although the output is prettier to read). If we want better test labels I will have to debug some more.


tests/run-perfreg.sh
```diff
@@ -3,51 +3,52 @@
...
+runTest "/bzt-configs/tests/scenarios/performance/performanceRegression.yaml"
+
+PERF_REG_RESULTS="${OUTPUT_DIR}/perf-reg-results"
+echo "Copying test results xml files to ${PERF_REG_RESULTS}"
+mkdir ${PERF_REG_RESULTS}
+kubectl -n ${CLUSTER_NAMESPACE} cp "${POD_NAME}:/bzt-configs/tests/results" "${PERF_REG_RESULTS}"
+
+# First Check to see if python3 is available for use with xmlCombiner.py script. If not first install python3.8 using yum. 
+pythonVersion=$(python3 --version 2>/dev/null)
+pythonBinary=python3
```
csandersdev, line 40: Seems like you would want to set pythonBinary above pythonVersion and then use it in the RHS of the pythonVersion expression.

tests/run-perfreg.sh
```diff
@@ -3,51 +3,52 @@
...
+if [ -z "${pythonVersion}" ]
+then
+  echo "Python check output: '${pythonVersion}'"
+  echo "Missing python3, installing..."
```
csandersdev, line 44: Use ``${pythonBinary}`` in the echo message?

tests/run-perfreg.sh
```diff
@@ -3,51 +3,52 @@
...
+  sudo yum -y install python38
+  which python3.8
+  python3.8 --version
+  pythonBinary=python3.8
+else 
+  echo "Found 'python3' binary"
```
csandersdev, line 50: Use ``${pythonBinary}`` in the echo message?

tests/src/main/python/TestDriver_MeasurePerformance.py
```diff
@@ -0,0 +1,52 @@
...
+
+def setup():
+    os.chdir(baseDir)
+    tests = list()
+    with open(testFile) as f:
+        data = json.load(f)
+        testValues = data['tests']
+        for testName in testValues:
+            tests.append((testValues[testName]['jsonMeasureConfigurationFile'], testValues[testName]['targets'], testName))
+    return tests
```
csandersdev, line 20: My python has never been great, but I find it odd that this method exists outside of the class that uses it. Is there a reason?

csandersdev (approved):
A few minor points, but nothing blocking.

tests/src/main/python/TestDriver_MeasurePerformance.py
msargentibm: Nope. Just copy pasted one of the other python scripts and modified it

tests/run-perfreg.sh
msargentibm: Solid suggestion.  All of the python binary bits was also me being lazy copy/pasting. But I'm happy to  clean things  up  a bit.

dkwasny-ibm (approved):
Worked with Matt to figure out how to make the test reports be a bit more indicative of what test cases are what.  Approving under the assumption he will update the cases accordingly.

tests/src/main/python/TestDriver_MeasurePerformance.py
msargentibm: Moving setup inside the class might be possible, but I'd have to learn more about pytest.mark.parametrize to know  how to do it correctly.

I don't think you can move it inside the class  and leverage the `@staticmethod` notation. I'm also not sure if moving setup to inside the class will work (with or without the `self` parameter).

jaigoradia (approved):
LGTM

PR #88, opened 2021-03-22
Prevent implicit use of transitive dependencies

jillrdoty:
This looks like a ton of pom changes (and it kind of is) - for the record, I did not remove slf4j-simple from any pom, just made sure it was a runtime dependency.

I also re-organized some of the dependencies to group by scope (default, runtime, test).

The only real interesting change is in the parent pom where I fail the build on implicit transitive dependencies here: https://github.com/Alvearie/quality-measure-and-cohort-service/pull/88/files#diff-e3290d13c090ff9b4372aa936379ba6c12707692e20aa505989167fc34c66511R1029

Finally, I realized that the javadoc warnings weren't failing the build due to an old version of the javadoc plugin, so I updated that and fixed a few things that had snuck in since I last fixed that.


cohort-parent/pom.xml
```diff
@@ -492,6 +561,47 @@
 				<artifactId>cxf-rt-transports-http</artifactId>
 				<version>3.3.8</version>
 			</dependency>
+			<dependency>
+				<groupId>com.fasterxml.jackson.core</groupId>
+				<artifactId>jackson-annotations</artifactId>
+				<version>2.11.2</version>
```
csandersdev, line 567: Looks like we repeat this version a few times, so it is probably worth having a property for it.

cohort-parent/pom.xml
```diff
@@ -910,6 +1020,24 @@
...
+							<failOnWarning>true</failOnWarning>
+							<ignoredUsedUndeclaredDependencies>
+								<!-- Excluding this because the JAXB classes are provided by the JRE as of Java 8, but there are potentially other classes in jakarta that we don't want to wholesale exclude -->
+								jakarta.xml.bind:jakarta.xml.bind-api
+							</ignoredUsedUndeclaredDependencies>
+							
+							<ignoredUnusedDeclaredDependencies> 
+								<ignoredUnusedDeclaredDependency>ca.uhn.hapi.fhir:hapi-fhir-structures-r4</ignoredUnusedDeclaredDependency>	
+								<ignoredUnusedDeclaredDependency>org.slf4j:slf4j-simple</ignoredUnusedDeclaredDependency>			
+							</ignoredUnusedDeclaredDependencies>
```
csandersdev, line 1038: Do we still need this ignoreUnusedDeclaredDependency after you added runtime to slf4j-simple dependency in cohort-cli/pom.xml?

cohort-parent/pom.xml
```diff
@@ -492,6 +561,47 @@
...
+			</dependency>
+			<dependency>
+				<groupId>commons-collections</groupId>
+				<artifactId>commons-collections</artifactId>
+				<version>3.2.2</version>
+			</dependency>
+			<dependency>
+				<groupId>io.swagger</groupId>
+				<artifactId>swagger-annotations</artifactId>
+				<version>1.6.2</version>
```
csandersdev, line 587: Another place here where version is repeated and could be a property

cohort-util/pom.xml
```diff
@@ -21,6 +21,38 @@
 
 	<dependencies>
 
+		<dependency>
+			<groupId>org.apache.poi</groupId>
+			<artifactId>poi-ooxml</artifactId>
+			<version>4.1.2</version>
```
csandersdev, line 27: Repeated version could be made into a property

csandersdev (commented):
I called out a few places where related dependencies that share a version have that version repeated in the POM and could instead be made into a shared property. Not blocking, but nice to have if/when we need to change those versions in the future.

You may know this, but changing the order of dependencies in Maven has a very real impact on the order the JARs appear in the classpath. When multiple versions of a dependency are requested, Maven will choose the version that is closest to the top of the root pom. Because of this behavior, it is generally a bad idea to just reorder them for presentation. There have been a number of times where I've specifically laid out dependencies in a particular order to force Maven to use my version vs. a transitive version. Dependency management is a better way to do this and we've done a pretty good job of using it, so I doubt that it has an impact here. Mostly, this is an FYI for the future.

cohort-parent/pom.xml
jillrdoty: Yup, the plugin isn't intelligent enough to realize it's actually used at runtime.

cohort-parent/pom.xml
jillrdoty: Thanks for calling this out, I meant to address this since we do have a jackson property that was set to something else 😬 .

deekim (approved):

PR #90, opened 2021-03-25
Wffhcohort-233

shelleyMurphy:
Adding the endpoint for creating and updating value sets to the liberty server.


cohort-engine-api-web/pom.xml
```diff
@@ -273,5 +279,49 @@
...
+			<type>jar</type>
+			<scope>compile</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>ch.qos.logback</groupId>
+			<artifactId>logback-core</artifactId>
+			<version>1.2.3</version>
+			<scope>compile</scope>
+		</dependency>
```
csandersdev, line 303: We don't want a compile-time dependency on a specific log implementation. This could be "test" scope as needed, but I've been using slf4j-simple as the test binding in other projects.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -279,5 +287,62 @@ public Response getMeasureParametersById(@Context HttpHeaders httpHeaders,
...
+								@ApiParam(value = "Multipart form request containing value set spreadsheets") IMultipartBody multipartBody
+								) {
+
+		String[] authParts = FHIRRestUtils.parseAuthenticationHeaderInfo(httpHeaders);
+		IGenericClient terminologyClient = FHIRRestUtils.getFHIRClient(fhirEndpoint, authParts[0], authParts[1], fhirTenantIdHeader, fhirTenantId, fhirDataSourceIdHeader, fhirDataSourceId);
+
+		try {
+			IAttachment valueSetAttachment = multipartBody.getAttachment(VALUE_SET_PART);
+			if (valueSetAttachment == null) {
+				throw new IllegalArgumentException(String.format("Missing '%s' MIME attachment", "temp"));
```
csandersdev, line 323: Shouldn't "temp" here be VALUE_SET_PART?

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -279,5 +287,62 @@ public Response getMeasureParametersById(@Context HttpHeaders httpHeaders,
...
+			ValueSetArtifact artifact;
+			try (InputStream is = valueSetAttachment.getDataHandler().getInputStream()) {
+				artifact = ValueSetUtil.createArtifact(is);
+			}
+			ValueSetUtil.validateArtifact(artifact);
+
+			boolean valueSetExists = ValueSetUtil.doesValueSetExist(terminologyClient, artifact);
+			if(valueSetExists && !updateIfExists){
+				//todo look up appropriate response
+				return Response.status(490).build();
```
csandersdev, line 335: Why HTTP 490? That isn't on the [standard list](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html) I'm familiar with. Maybe 403 Forbidden? Also, it would be preferable to use the Response.Status.XXX constants here for readability.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -279,5 +287,62 @@ public Response getMeasureParametersById(@Context HttpHeaders httpHeaders,
...
+								@ApiParam(value = CohortEngineRestHandler.FHIR_ENDPOINT_DESC, required = true, defaultValue = CohortEngineRestHandler.DEFAULT_FHIR_URL) @QueryParam("fhir_server_rest_endpoint") String fhirEndpoint,
+								@ApiParam(value = CohortEngineRestHandler.FHIR_TENANT_ID_DESC, required = true, defaultValue = "default") @QueryParam("fhir_server_tenant_id") String fhirTenantId,
+								@ApiParam(value = CohortEngineRestHandler.FHIR_TENANT_HEADER_DESC, defaultValue = IBMFhirServerConfig.DEFAULT_TENANT_ID_HEADER) @QueryParam("fhir_server_tenant_id_header") String fhirTenantIdHeader,
+								@ApiParam(value = CohortEngineRestHandler.FHIR_DS_HEADER_DESC, defaultValue = IBMFhirServerConfig.DEFAULT_DATASOURCE_ID_HEADER) @QueryParam("fhir_data_source_id_header") String fhirDataSourceIdHeader,
+								@ApiParam(value = CohortEngineRestHandler.FHIR_DS_ID_DESC) @QueryParam("fhir_data_source_id") String fhirDataSourceId,
+								//todo default to false
+								@QueryParam("updateIfExists") boolean updateIfExists,
+								//todo can assume only need one file at a time
+								@ApiParam(value = "Multipart form request containing value set spreadsheets") IMultipartBody multipartBody
```
csandersdev, line 314: If we wanted to support multiple files at a time, it would be pretty easy to loop through the attachments on the multipart request and check the Content-Type as .xlsx or something like that. It wouldn't need an additional API parameter.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -279,5 +287,62 @@ public Response getMeasureParametersById(@Context HttpHeaders httpHeaders,
...
+								) {
+
+		String[] authParts = FHIRRestUtils.parseAuthenticationHeaderInfo(httpHeaders);
+		IGenericClient terminologyClient = FHIRRestUtils.getFHIRClient(fhirEndpoint, authParts[0], authParts[1], fhirTenantIdHeader, fhirTenantId, fhirDataSourceIdHeader, fhirDataSourceId);
```
csandersdev, line 318: We should have the following code block at the beginning of the method...

```java
			// Perform api setup
			Response errorResponse = ServiceBaseUtility.apiSetup(version, logger, methodName);
			if(errorResponse != null) {
				return errorResponse;
			}
```
and then this at the end...
```java
		} finally {
			// Perform api cleanup
			Response errorResponse = ServiceBaseUtility.apiCleanup(logger, methodName);
			if( errorResponse != null ) {
				response = errorResponse;
			}
		}
```

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
```diff
@@ -65,10 +77,37 @@
...
 	public static void setUp() throws Exception {
 		// Instantiate service location
 		instantiateServiceLocation();
+
+		if( System.getProperty("test.host") == null ) {
+			System.setProperty("test.host", "localhost");
+		}
+
+		if(  System.getProperty("test.httpPort") == null ) {
+			System.setProperty("test.httpPort", "9089");
```
csandersdev, line 92: I had this set as 9080 on my side. I don't really care what we decide on as the default, but it should be agreed upon across the team. 9080 is the default liberty port, but that obviously means that other services, such as IBM FHIR, want to also live there. There is a decent reason we might choose your port instead, but it is worth a quick conversation with the wider team.

cohort-parent/pom.xml
```diff
@@ -1037,16 +1037,19 @@
 							<failOnWarning>true</failOnWarning>
 							<ignoredUsedUndeclaredDependencies>
 								<!-- Excluding this because the JAXB classes are provided by the JRE as of Java 8, but there are potentially other classes in jakarta that we don't want to wholesale exclude -->
-								jakarta.xml.bind:jakarta.xml.bind-api
+								<ignoredUsedUndeclaredDependency>jakarta.xml.bind:jakarta.xml.bind-api</ignoredUsedUndeclaredDependency>
+								<ignoredUsedUndeclaredDependency>com.sun.activation:jakarta.activation</ignoredUsedUndeclaredDependency>
```
csandersdev, line 1041: In my PR (#178), I did a bunch of work to try to normalize our javax.activation dependency. Because of different artifactId and groupId in transitive dependencies, we were actually pulling in several copies of the activation API and implementation JARs. I normalized onto the following....

```xml
		<dependency>
			<groupId>jakarta.activation</groupId>
		    	<artifactId>jakarta.activation-api</artifactId>
		    	<version>1.2.2</version>
		</dependency>
```

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
```diff
@@ -27,6 +27,22 @@
...
+				throw new IllegalArgumentException("Identifier must be supplied, ensure that either the OID or the ID field is filled in");
+			}
+			if(valueSetArtifact.getFhirResource().getVersion() == null || valueSetArtifact.getFhirResource().getVersion().equals("")){
+				throw new IllegalArgumentException("Value Set Version must be supplied");
+			}
+			if(valueSetArtifact.getFhirResource().getCompose().getInclude() == null || valueSetArtifact.getFhirResource().getCompose().getInclude().size() == 0){
+				throw new IllegalArgumentException("Value set must include codes but no codes were included.");
+			}
+	}
+
```
csandersdev, line 45: Because there was a lot of need for validation in POJOs, I introduce a dependency on javax.validation in my PR and rely on mostly annotations to define the validation constraints for model objects. Nothing wrong with what you've done here, but maybe another conversation to have with the wider team about validation and making sure we are consistent across code paths.

csandersdev (changes requested):
Looks pretty good. I did leave a few comments where there are some definite changes I'd like to see. Beyond those, I noted that you have some unresolved todos that we will obviously want to cleanup before this PR goes to a wider audience.

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
```diff
@@ -27,6 +27,22 @@
...
+			if(valueSetArtifact.getFhirResource().getId() == null || valueSetArtifact.getFhirResource().getId().equals("")){
+				throw new IllegalArgumentException("Identifier must be supplied, ensure that either the OID or the ID field is filled in");
+			}
+			if(valueSetArtifact.getFhirResource().getVersion() == null || valueSetArtifact.getFhirResource().getVersion().equals("")){
+				throw new IllegalArgumentException("Value Set Version must be supplied");
+			}
+			if(valueSetArtifact.getFhirResource().getCompose().getInclude() == null || valueSetArtifact.getFhirResource().getCompose().getInclude().size() == 0){
+				throw new IllegalArgumentException("Value set must include codes but no codes were included.");
+			}
+	}
```
tabishop, line 44: Another set of validation checks that might be good to have is to make sure there are no SQL characters in any of the fields. AppScan on Cloud will flag places where we take input parameters and don't do validation on them to avoid SQL Injection attacks. I'm not sure how FHIR handles these fields, or if they would even be getting to a place where any SQL contained in them could be used, but thought I would mention in case it applies.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
tabishop: FYI, since it was decided that the taurus framework will be used to test these classes, the DefaultVT won't get run out on the pod and these are essentially integration tests that we can run during development time. They don't get run during the build either. 

The way I call these form eclipse is with a Launcher configuration that always passes in the following as VM args, -Dtest.host=localhost -Dtest.httpPort=9080 -Dtest.httpSslPort=9443 -Dtest.contextRoot=services/cohort -Dtest.enabledDarkFeatures=all

That being said, I would agree 9080 would be the default and would normally be overriden using the jvm arg.

tabishop (changes requested):
Couple of comments, but think we should change the getting of the FHIR client to allow OAUTH (unless there was explicit discussion of not allowing it which I was not aware of).

cohort-engine-api-web/pom.xml
shelleyMurphy: I had to add this for the dependency analyzer, I'll try removing it again and see what pops up.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
shelleyMurphy: Oh no, I managed to lose a bunch of changes when I squashed! I'm gonna try to revert this file from my IDE, hopefully that resolves this stuff.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
shelleyMurphy: 409 is what Jeff asked for, and it was that before I messed the commits up, but I'm happy using .CONFLICT instead

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
shelleyMurphy: My 9080 port was already being used when I tried to start my testing, so I went with a less commonly used one. I'm fine with making it 9080 for what's committed, but that's the logic there at least.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
shelleyMurphy: I'm fine to do that, though the feedback I've gotten right now is that there's not going to be a use case for the UI to upload multiple value sets at once.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
shelleyMurphy: I did in the pre-disaster era, should be back now.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -279,5 +287,62 @@ public Response getMeasureParametersById(@Context HttpHeaders httpHeaders,
...
+	@Consumes(MediaType.MULTIPART_FORM_DATA)
+	@Produces({ MediaType.APPLICATION_JSON })
+	@ApiResponses(value = {
+			@ApiResponse(code = 200, message = "Successful Operation", response = String.class),
+			//todo new response code for already exists
+			@ApiResponse(code = 400, message = "Bad Request", response = ServiceErrorList.class),
+			@ApiResponse(code = 500, message = "Server Error", response = ServiceErrorList.class)
+	})
+	@ApiOperation(value = "Insert a new value set to the fhir server or, if it already exists, update it in place", authorizations = {@Authorization(value = "BasicAuth") })
+	public Response putValueSet(@Context HttpHeaders httpHeaders,
```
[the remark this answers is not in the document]
shelleyMurphy: I think we do want it, and my original thought was to let Corey merge his PR back first so we wouldn't have conflicts of both adding those constants, but I think it's better for me to just checkout those files and implement it now.

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
shelleyMurphy: Does that kind of validation not happen in the fhir client itself? I would have assumed as much, since it's the one that actually makes the insertions. Do we do this kind of validation other places that I can see?

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
```diff
@@ -219,9 +262,70 @@ public void testXMLUnitAssertXPathValues() throws Exception {
...
+				.queryParam("updateIfExists", false)
+				.multiPart(CohortEngineRestHandler.VALUE_SET_PART, new File("src/test/resources/2.16.840.1.113762.1.4.1114.7.xlsx"));
+
+		ValidatableResponse response = request.post(RESOURCE, getServiceVersion()).then();
+		runSuccessValidation(response, ContentType.JSON, HttpStatus.SC_OK);
+	}
+
+	@Test
+	public void testValueSetAlreadyExists(){
+		testValueSetUpload();
```
csandersdev, line 284: Does this rely on being run after the previous testValueSetUpload() method in order to be successful? If so, consider rewriting the test so that it is completely self sufficient. For example, you could call the upload twice in this method. Test order is not guaranteed by the test runners.

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
```diff
@@ -97,30 +114,43 @@ else if (inCodesSection) {
...
+				outcome = client.update().resource(client.getFhirContext().newJsonParser().encodeResourceToString(valueSetArtifact.getFhirResource()))
+						.conditional().where(ValueSet.URL.matches().value(valueSetArtifact.getUrl())).execute();
+			}
+			else{
+				return null;
 			}
 		}
+		else {
+			outcome = client.create().resource(client.getFhirContext().newJsonParser().encodeResourceToString(valueSetArtifact.getFhirResource())).execute();
+		}
```
csandersdev, line 150: There is a race condition here between when the search method returns and the create/update method runs. During execution, we might try to create a resource that shows up after the search completes or update a resource that has changed since we last saw it.  FHIR has [capabilities](https://www.hl7.org/fhir/http.html#concurrency) built in to handle conditional creates and concurrency issues and there are [HAPI client examples](https://hapifhir.io/hapi-fhir/docs/client/examples.html) available for how to use them. It looks something like this...

``` java
bundle.addEntry()
   .setFullUrl(patient.getIdElement().getValue())
   .setResource(patient)
   .getRequest()
      .setUrl("Patient")
      .setIfNoneExist("identifier=http://acme.org/mrns|12345")
      .setMethod(Bundle.HTTPVerb.POST);
```

The odds here of the race condition mattering are pretty slim. I am *not* requesting changes to your PR. I just want to document this because it is something I thought about when I created the MeasureImporter. In the end, I just did a PUT for everything in my bundle because PUT in IBM FHIR will do [update as create](https://www.hl7.org/fhir/http.html#upsert) if the resource isn't there (as long as you provide a Resource.id in your resource). I didn't have a requirement for the "block if exists" behavior.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
```diff
@@ -219,9 +262,70 @@ public void testXMLUnitAssertXPathValues() throws Exception {
...
 	protected Logger getLogger() {
 		return logger;
 	}
 
+
+	@After
+	public void cleanUp(){
+
+		IGenericClient terminologyClient = FHIRRestUtils.getFHIRClient(dataServerConfig.getEndpoint(), dataServerConfig.getUser(), dataServerConfig.getPassword(), null, null, null, null);
+		ValueSetUtil.deleteValueSet(terminologyClient, "http://cts.nlm.nih.gov/fhir/ValueSet/testValueSet");
```
csandersdev, line 328: This makes a lot of sense to reset the FHIR server to a known state, but still not sure how those earlier update tests work unless they are relying on data created by the first create test.

csandersdev (approved):
I have some thoughts on the tests and a few notes about the design of the import method, but neither are blocking.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
shelleyMurphy: The name of the upload is misleading, but the testValueSetUpload does not have the test annotation and does not run on its own. The two tests that depend on it run that method, which both uploads and confirms that the response is good, and the cleanup method deletes the value set each time. I would be fine with either rewording the upload method or combining the three tests into one, I just wanted to be clear that there currently isn't an order dependency.

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
shelleyMurphy: Thanks for this, definitely has given me something to think about.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
```diff
@@ -219,9 +262,70 @@ public void testXMLUnitAssertXPathValues() throws Exception {
...
+				.queryParam("updateIfExists", false)
+				.multiPart(CohortEngineRestHandler.VALUE_SET_PART, new File("src/test/resources/2.16.840.1.113762.1.4.1114.7.xlsx"));
+
+		ValidatableResponse response = request.post(RESOURCE, getServiceVersion()).then();
+		runSuccessValidation(response, ContentType.JSON, 409);
+	}
+
+	@Test
+	public void testValueSetOverride(){
+		testValueSetUpload();
```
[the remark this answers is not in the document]
csandersdev: I realize looking back on this that I highlighted the part of the code that actually answered my question. You are doing it right. No action needed.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
csandersdev: I realize looking back on this that I highlighted the part of the code that actually answered my question. You are doing it right. No action needed.

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
```diff
@@ -97,30 +114,43 @@ else if (inCodesSection) {
...
-					valueSetArtifact.setId(outcome.getId().getIdPart());
-				}
-			} else {
-				MethodOutcome outcome = client.create().resource(client.getFhirContext().newJsonParser().encodeResourceToString(valueSetArtifact.getResource())).execute();
-				valueSetArtifact.setId(outcome.getId().getIdPart());
+		Bundle bundle = client.search().forResource(ValueSet.class).where(ValueSet.URL.matches().value(valueSetArtifact.getUrl()))
+				.returnBundle(Bundle.class).execute();
+		MethodOutcome outcome;
+		if(bundle.getEntry().size() > 0){
+			String[] url = bundle.getEntry().get(0).getFullUrl().split("/");
```
msargentibm, line 138: Sanity checking myself:
Bundle entries have full url and the actual resource entry. The last part of the full url will always match the resource's id. Is that right?

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
```diff
@@ -376,9 +383,71 @@ public void testXMLUnitAssertXPathValues() throws Exception {
...
+		byte[] encodedAuth = Base64.encodeBase64(auth.getBytes(StandardCharsets.ISO_8859_1));
+
+		RequestSpecification request = buildBaseRequest(new Headers(new Header("Authorization", "Basic " + new String(encodedAuth))))
+				.param("version", ServiceBuildConstants.DATE)
+				.queryParam("fhir_server_rest_endpoint", dataServerConfig.getEndpoint())
+				.queryParam("updateIfExists", false)
+				.multiPart(CohortEngineRestHandler.VALUE_SET_PART, new File("src/test/resources/2.16.840.1.113762.1.4.1114.7.xlsx"));
+
+		ValidatableResponse response = request.post(RESOURCE, getServiceVersion()).then();
+		runSuccessValidation(response, ContentType.JSON, 409);
```
msargentibm, line 419: Can `409` be `HttpStatus.SC_CONFLICT` here?

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -152,16 +156,17 @@ public CohortEngineRestHandler() {
...
-			@ApiResponse(code = 400, message = "Bad Request", response = ServiceErrorList.class), 
-			@ApiResponse(code = 500, message = "Server Error", response = ServiceErrorList.class) 
+			@ApiResponse(code = 400, message = "Bad Request", response = ServiceErrorList.class),
+			@ApiResponse(code = 500, message = "Server Error", response = ServiceErrorList.class)
 	})
 	public Response evaluateMeasure(@Context HttpServletRequest request,
+
 			@ApiParam(value = ServiceBaseConstants.MINOR_VERSION_DESCRIPTION, required = true, defaultValue = ServiceBuildConstants.DATE) @QueryParam("version") String version,
-			@ApiParam(hidden = true, type="file", required=true) IMultipartBody multipartBody) {		
+			@ApiParam(value = "patients and the measures to run", required = true) IMultipartBody multipartBody) {
```
msargentibm, line 167: Wanted to double check if this change was intentional since it's in `evaluateMeasure`

msargentibm (approved):
Nice work.

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
shelleyMurphy: Yup!

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
shelleyMurphy: Definitely, will change

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
shelleyMurphy: Good catch, I think this was something Corey and I were messing with yesterday and I messed with weirdly.

cohort-engine-api-web/src/test/java/com/ibm/cohort/engine/api/service/vt/DefaultVT.java
msargentibm: Oh, I thought that constant would work since it's from the same place as `HttpStatus.SC_OK` used in some of the other tests.

shelleyMurphy: I misread your import, that does work, I'll make that change.

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
dkwasny-ibm: @tabishop I'm not sure how feasible it is to scan every single field in a complex POJO like this.  Any new fields to either `ValueSetArtifact` or HAPI's `ValueSet` object would render the method obsolete.  Would AppScan be able to tell if we instead attempt to validate against the JSON string that we ultimately send the FHIR server?

Also, do we perform SQL injection tests during our measure/library upload pipeline as well?

dkwasny-ibm (changes requested):
Nothing crazy, just a small suggestion to keep our dependency tree down if possible.

dkwasny-ibm (approved):

cohort-util/src/main/java/com/ibm/cohort/valueset/ValueSetUtil.java
tabishop: I think checking the JSON string would probably be ok. I don't have a lot of specifics on how to code the checks that the AppScan tool is flagging. It basically says "you should be doing validation on these input parameters to avoid SQL injection attacks" but it also flags a lot of things that aren't actually problems either. Just wanted to raise it so there is awareness that we may get dinged on it if we don't have an answer at where this is being done.

tabishop (approved):
Approved, but we should at least try injecting some SQL into the input parms to see what happens.

PR #91, opened 2021-03-25
Add CT performance tests

msargentibm:
The primary part of this PR was adding performance tests using the `tests/run-performancetests.sh` file. It leverages `tests/scenarios/performance/performanceCTScenarios.yaml` and  `tests/src/main/resources/performance/one-thousand-patients-five-measures.json` to run tests in a similar manner as the performance regression tests. Some other changes of note:

- I turned off chaos testing since the step in the CT toolchain didn't appear to actually function properly and only served to inflate CT  the toolchain runtime. We should experiment with turning it on again at some  point in the  future.
- I added some light verification rules to the Python  script used for performance testing (used  for both  performance and performance regression testing).
- `kubectl exec` was timing out for long-running performance tests before Taurus actually finished running. I had to add a wait and retry loop alongside the normal `kubectl cp` command  to prevent the toolchain from moving on without the expected results files.


deekim (approved):
Matt walked me through the changeset yesterday.
LGTM.

dkwasny-ibm (approved):

PR #93, opened 2021-03-29
add cos integration

deekim:
This change's primary purpose is to add the `COS` dependency to access it through the `measureCLI` code path.
Because we are still working through implementing the persistence cache layer, the code "doesn't do anything" yet.

I added a `CosExampleCLI` for testing and example purposes, 
which I can revert if people don't think that should be merged back.

## additional notes
I spent a bit of extra time (08a91bd) trying to support the concept of "dynamic COS client configuration" 
but ultimately decided to revert the commit because I thought it wasn't worth the added complexity. 
Let me know if you feel differently, and I will _revert the revert_.

```bash
# example usage
java -classpath code.jar com.ibm.cohort.cli.MeasureCLI \
-Ccos.request.timeout=5000 -Ccos.tcp.keep.alive=true ...
```

## references
- [WFFHCOHORT-307](https://jira.wh-sdlc.watson-health.ibm.com/browse/WFFHCOHORT-307)


cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosDao.java
```diff
@@ -0,0 +1,73 @@
...
+
+	public String getObjectAsString(String bucket, String key) throws IOException {
+		logger.trace("Retrieving object from bucket `{}` with key: `{}`", bucket, key);
+		try (
+				S3Object object = cosClient.getObject(new GetObjectRequest(bucket, key));
+				InputStream stream = object.getObjectContent()
+		) {
+			return IOUtils.toString(stream, StandardCharsets.UTF_8.name());
+		} catch (IOException e) {
+			logger.error("Error reading file: `{}`", key, e);
```
msargentibm, line 42: Would calling out the bucket here be useful too in case log level is set for more severe than `trace`?

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
```diff
@@ -89,6 +89,9 @@
 		@Parameter(names = { "--disable-retrieve-cache" }, description = "Disable the use of the retrieve cache.")
 		private boolean disableRetrieveCache = false;
 
+//		@ParametersDelegate
```
msargentibm, line 92: Is this commented code prep work for the next task?

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+	@Parameter(names = { "-k", "--cos-api-key" }, description = "COS API key" )
+	private String apiKey;
+
+	@Parameter(names = { "-s", "--cos-id" }, description = "COS service instance id" )
+	private String serviceInstanceId;
+
+	@Parameter(names = { "-u", "--cos-url" }, description = "COS endpoint url" )
+	private String endpointUrl = "https://s3.us-south.cloud-object-storage.appdomain.cloud";
+
+	@Parameter(names = { "-l", "--cos-location" }, description = "COS Location" )
```
msargentibm, line 26: Will these parameters eventually be hooked in to the MeasureCLI? Wondering if we need to worry about collisions with parameters in MeasureCLI/BaseCLI (ex: https://github.com/Alvearie/quality-measure-and-cohort-service/blob/51762d96ee5e3bc3393701d2a2ddc5afdf3abbb1/cohort-cli/src/main/java/com/ibm/cohort/cli/CohortCLI.java#L58).

msargentibm (approved):

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosDao.java
deekim: Thanks, good idea.

going with:
```suggestion
			logger.error("Error retrieving object from bucket `{}` with key: `{}`", bucket, key, e);
```

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosConfiguration.java
deekim: Correct. 
`JCommander` does throw an `Exception` in that case, but it's at `runtime`.
I can look into adding build-time validation, but I believe the change would not be trivial 
(probably a 50 to 100 line change).

I'll leave this comment unresolved, to see if anyone else wanted to chime in.

cohort-cli/src/main/java/com/ibm/cohort/cli/MeasureCLI.java
deekim: Yes. It was more so to provide an example of usage (0e84a19).
I can revert the commit if there are objects to having it in the codebase,
but the uncommented version will likely find its way into code at some point 😉

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosConfiguration.java
deekim: @rickjstevens, those are all great questions that are outside the scope of this ticket.
@dkwasny-ibm are talking through some of those points that you brought up now,
but I'll make sure we keep your comments in mind during design.

cohort-engine/pom.xml
```diff
@@ -108,7 +108,22 @@
...
+
+		<dependency>
+			<groupId>com.ibm.cos</groupId>
+			<artifactId>ibm-cos-java-sdk-core</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>com.ibm.cos</groupId>
+			<artifactId>ibm-cos-java-sdk-s3</artifactId>
+		</dependency>
```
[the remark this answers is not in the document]
dkwasny-ibm: After thinking about it a bit, I think I agree.  We could have the COS integration work in a separate project and have something implement the `RetrieveCacheContext` as a way to integrate the changes into the `cohort-engine` codebase.
That way users can import the project, and use the respective implementation if they desire.
Configuration will be a little more complex, but should be fine especially with the pattern Caffeine has laid out for us.

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.ibm.cohort.engine.cos;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.beust.jcommander.DynamicParameter;
+import com.beust.jcommander.Parameter;
+
+public class CosConfiguration {
```
dkwasny-ibm, line 15: Per my earlier comment, I'm not sure if COS specific information has a place here.
That does mean tricky things for how we handle configuration of the COS information.
We may want to use a combination of config files (like Caffeine) and environment variables instead.

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosDao.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.ibm.cloud.objectstorage.services.s3.AmazonS3;
+import com.ibm.cloud.objectstorage.services.s3.model.Bucket;
+import com.ibm.cloud.objectstorage.services.s3.model.GetObjectRequest;
+import com.ibm.cloud.objectstorage.services.s3.model.S3Object;
+
+public class CosDao {
```
dkwasny-ibm, line 24: This class might be a little "cart before the horse", but I'm OK keeping it around (but in a different project) to see if we get value out of it.

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosDaoFactory.java
```diff
@@ -0,0 +1,47 @@
...
+		);
+
+		return new CosDao(cosClient);
+	}
+
+	private static AmazonS3 createClient(String apiKey, String serviceInstanceId, String endpointUrl, String location) {
+		AWSCredentials credentials = new BasicIBMOAuthCredentials(apiKey, serviceInstanceId);
+		ClientConfiguration clientConfig = new ClientConfiguration()
+				.withRequestTimeout(5_000)
+				.withTcpKeepAlive(true);
```
dkwasny-ibm, line 37: We should be able to move these to a config file or something.
Nothing immediately needed, but just want to log my thoughts.

dkwasny-ibm (changes requested):
I think the key things worth discussing here are:
1. Where the code lives
2. How it is configured

I have my thoughts expressed in my comments, but I would like to hear your opinions as well.

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosDao.java
deekim: Correct. The intent was to have this code be the backing implementation to a #future class called `CosCacheProvider`.

cohort-engine/src/main/java/com/ibm/cohort/engine/cos/CosDaoFactory.java
deekim: @CoreySandersOH @dkwasny-ibm 
_re:_
> Where the code lives

After thinking about this a bit more, 
I do agree that it should be it's in a separate module for the same reasons you two are talking about.

I guess my question would be should I also move parts of the caching interface code as well at that point?

deekim:
@CoreySandersOH @dkwasny-ibm 
_re:_
> Where the code lives

After thinking about this a bit more, 
I do agree that it should be it's in a separate module for the same reasons you two are talking about.

I guess my question would be should I also move parts of the caching interface code as well at that point?

dkwasny-ibm:
> @CoreySandersOH @dkwasny-ibm
> _re:_
> 
> > Where the code lives
> 
> After thinking about this a bit more, I do agree that it should be in a separate module
> for the same reasons you two are talking about.
> 
> I guess my question would be:
> should I also move parts of the caching interface code as well at that point?

I think in a "purist" world, the caching interface would be in its own project that gets imported into both the engine and the COS caching layer.  I don't see a need to go that far with this release though.  Having the caching layer show up in the engine project along with the simple JCache implementation doesn't feel all that bad.

The main reason for separating the code is to prevent dependency bloat.

deekim:
08cb335 moves the primary chunk of `cos` access code to its own module and 
a6031bb changes the configuration pattern to be more in line with how we pass `fhir` server information via the `cli` 
(via a `json` file reference)

csandersdev (commented):
Nitpicking a bit here, but if I were to pick a name for the new COS module, I would have picked cohort-cache-cos with the expectation that there might be a cohort-cache-rdms module later (or anything really) and I would want them to sort together. I would also expect possibly a cohort-cache module to exist at some point that defines the required interfaces for each of those sub modules. Again, they would all group together.

Looking at the code you have written so far, is it right that these are all infrastructure classes for providing an interface to COS and the actual work to implement the cache is coming later in this same project? With just what is there right now, it feels more like cohort-dao-cos than cohort-cache-cos. Not sure we would want to break the dao and the cache implementations up into separate projects, but it feels like we could if there was an expectation that other parts of the system might want the same DAO layer.

I got onto the previous point about dao vs. cache because the package names didn't seem to match the module name. I was expecting something like com.ibm.cohort.engine.cache.cos vs. com.ibm.cohort.engine.cos, Based on the functionality that is there, I don't mind the exist package name, but it took me some time to grok what was going on and if it were me I probably would put service or dao in the package name before cos.

cohort-cos-cache/pom.xml
```diff
@@ -0,0 +1,117 @@
...
+	<dependencies>
+		<dependency>
+			<groupId>com.ibm.cos</groupId>
+			<artifactId>ibm-cos-java-sdk-core</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>com.ibm.cos</groupId>
+			<artifactId>ibm-cos-java-sdk-s3</artifactId>
+		</dependency>
```
dkwasny-ibm, line 36: For my `cos-playground` project, I was able to get away with just a single dependency on `ibm-cos-java-sdk`.
In fact, you add that dependency to the BOM, but don't explicitly depend on it anywhere.

Does that pull in a bunch of extra dependencies you wanted to avoid?

cohort-cos-cache/pom.xml
```diff
@@ -0,0 +1,117 @@
...
+			<artifactId>commons-io</artifactId>
+		</dependency>
+
+		<!-- Runtime dependencies -->
+		
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-simple</artifactId>
+			<scope>runtime</scope>
+		</dependency>
```
dkwasny-ibm, line 54: Being a library, this module should not depend on a SLF4J implementation except for tests.

cohort-cos-cache/pom.xml
```diff
@@ -0,0 +1,117 @@
...
+				<version>3.0.0</version>
+				<executions>
+					<execution>
+						<goals>
+							<goal>java</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
```
dkwasny-ibm, line 114: I know this module doesn't need an uber jar to be created during the build.
I also don't _think_ we need the exec plugin either.

dkwasny-ibm (changes requested):
Just some thoughts on the `pom` file.

cohort-cos-cache/pom.xml
deekim:
```bash
Warning:  Used undeclared dependencies found:
1233
Warning:     com.ibm.cos:ibm-cos-java-sdk-s3:jar:2.9.0:compile
1234
Warning:  Unused declared dependencies found:
1235
Warning:     com.ibm.cos:ibm-cos-java-sdk:jar:2.9.0:compile
```

is the [error](https://github.com/Alvearie/quality-measure-and-cohort-service/runs/2291182385) I get if I do what you're suggesting. Do you know how to resolve that?

dkwasny-ibm:
> Nitpicking a bit here, but if I were to pick a name for the new COS module, I would have picked cohort-cache-cos with the expectation that there might be a cohort-cache-rdms module later (or anything really) and I would want them to sort together. I would also expect possibly a cohort-cache module to exist at some point that defines the required interfaces for each of those sub modules. Again, they would all group together.
> 
> Looking at the code you have written so far, is it right that these are all infrastructure classes for providing an interface to COS and the actual work to implement the cache is coming later in this same project? With just what is there right now, it feels more like cohort-dao-cos than cohort-cache-cos. Not sure we would want to break the dao and the cache implementations up into separate projects, but it feels like we could if there was an expectation that other parts of the system might want the same DAO layer.
> 
> I got onto the previous point about dao vs. cache because the package names didn't seem to match the module name. I was expecting something like com.ibm.cohort.engine.cache.cos vs. com.ibm.cohort.engine.cos, Based on the functionality that is there, I don't mind the exist package name, but it took me some time to grok what was going on and if it were me I probably would put service or dao in the package name before cos.

I'm personally OK with the proposed name change, but will leave it up to Dan.

My current thought is that the new module is a starting point for all the cache work.  Breaking out COS specific code for other parts of the project to use can come later when the need arises.

I'm also OK with the proposed package name change, and will leave the decision to Dan.
My view is that we can use the more generic `com.ibm.cohort.engine.cos` package in a new module if/when we extract usecase agnostic features of our COS integration.

cohort-cos-cache/pom.xml
dkwasny-ibm: Ah, I forgot about the unused dependency checker.  Yea do whatever you need to satisfy the checker.  Sorry for the confusion.

msargentibm (approved):
Seems like the requested changes are in place.

dkwasny-ibm (approved):

csandersdev (approved):
LGTM

shelleyMurphy (approved):

deekim:
I will leave the branch around to see if we want to revisit this next sprint but closing this PR for the time being.

PR #94, opened 2021-03-29
Update tests/README.md

csandersdev:
This update adds notes to the tests/README.md based on things that I encountered while trying to follow the instructions on a Windows laptop.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,7 +1,70 @@
-# Mac OS setup steps.
+# Test setup steps:
 
-## Install Taurus
+## Build and run the Docker image
 
+To build a docker image that  will run in the same cluster as the FHIR server you will use for testing, run:
+
+```bash
+# ensure you have the latest image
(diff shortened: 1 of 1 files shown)


msargentibm (approved):
What you have makes sense.

Did you ultimately just temporarily alter the couple of yaml/json files to point at your local FHIR configs, and then undo them before committing your real changes?

csandersdev:
> Did you ultimately just temporarily alter the couple of yaml/json files to point at your local FHIR configs, and then undo them before committing your real changes?

The only change I have locally to source files is that I've got our cloud-based FHIR server hostname, user, and password in the cohort-cli/config/local-fhir-config.json and tests/src/main/resources/config/local-fhir-config.json files. I would love a solution that enables me to have a single copy of my own version of those files without updating source, but it wasn't immediately obvious how to do that.

jaigoradia (approved):
Seems reasonable.

PR #102, opened 2021-04-08
Wffhcohort 306 -- Add timeout configuration options to FHIR server client configs

msargentibm:
I modified `FhirServerConfig.java` to include optional configurations to specify several client timeouts. The config options are used  in `DefaultFhirClientBuilder.java` when making the actual clients.

In order to support separate timeout settings for different FHIR clients, I had to use a separate FhirContext object for each client being created (we were previously using a single context for any  client we were creating). This lead to changes in the various CLI classes, CQLEngineWrapper,  and FHIRClientContext  to make sure FhirContext objects were not being reused across FHIR client creation.

I found testing to be a little tricky, but I tried to cover both using `DefaultFhirClientBuilder` directly  as well as using the `FHIRClientContext` class to make sure that timeout configurations were being handled correctly when making  a client. I was struggling to test `connectTimeout` and `connectionRequestTimeout`, so the WireMock tests focus on the `socketTimeout` configuration.

I added only minimal mocking for delayed requests for a Patient resource to `BaseFhirTest`. I didn't see adding delay to mocks as something we needed more  generally, but would be happy to revisit/refactor if folks think the functionality be useful to have for resources in general.


deekim (approved):

csandersdev (commented):
A few small questions. What happens if they set any of the timeouts to 0? Is that interpreted as unlimited? I imagine QA is going to do something like try to put a negative number there as well. We should understand what the behavior is going to be.

msargentibm:
@csandersdev, I've been having a tough time on locating the exact behaviors. It seems like for sockets 0 is infinite, negative is undefined. Connect might have similar behavior. Still looking around for connectionRequestTimeout.

FWIW, connectionRequestTimeout does specify "This setting typically does not need to be adjusted" so maybe I don't need to expose that in the first place?

For socket/connect timeouts I could add a check to make sure that the configs are >= 0 so that we don't have to deal with undefined behavior.

dkwasny-ibm (approved):

msargentibm:
Found an issue when using multiple configs with different timeouts set in the MeasureCLI. Going to have to try to refactor things a bit. Will  mark PR as [WIP] until I have something to review.

msargentibm:
Sorry to  re-request reviews, but I did have to tweak the code a bit to get multiple configs to respect independent timeout values. It's different enough that I wanted to give folks a chance to look things over again.

deekim (approved):

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/FHIRClientContext.java
```diff
@@ -6,13 +6,13 @@
...
 import com.ibm.cohort.fhir.client.config.DefaultFhirClientBuilderFactory;
 import com.ibm.cohort.fhir.client.config.FhirClientBuilder;
 import com.ibm.cohort.fhir.client.config.FhirClientBuilderFactory;
 import com.ibm.cohort.fhir.client.config.FhirServerConfig;
 
+import ca.uhn.fhir.rest.client.api.IGenericClient;
+
 /**
  * A wrapper around the different HAPI client connections required by the engine.
  *
```
csandersdev, line 18: You've done a good job updating the comments in other sections of the API, but I think I would like to also add some detail here that states... "FHIRClientContext objects are relatively expensive to create and should be reused as much as possible."

csandersdev (approved):
Added a small request for javadoc update because I'm still worried about users shooting themselves in the foot related to too many HAPI FhirContext objects being created. We want to encourage reuse of the clients and clientcontexts as much as possible. In that vein, I logged #WFFHCOHORT-358 requesting that we update CqlEngineWrapper to use the FhirClientContext stuff that Kwas created for the MeasureEvaluator path. That will help with consistency and clarity on how we expect our consumers to use the APIs. Approving either way. I like how seamlessly you were able to deal with the factory/builder issue.

PR #104, opened 2021-04-08
Wffhcohort 342 address open api validator tool issues

tabishop:
This PR addresses a bunch of issues from the openAPI validator tool as well as switches the passing of our REST fhir server connection details to use a file to be consistent between our APIs and other areas of the code such as the cli which already use this method.

Note that a few REST input parameter names were changed to follow the convention of the rest-service-framework which uses lower_snake_case on input parameters and lowerCamelCase for output parameters. The FHIR measure report object also uses lowerCamelCase and so we are using for output parameters to maintain consistency.


msargentibm (approved):
I went over these changes somewhat quickly. I would request at least another approval before merging.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortServiceExceptionMapper.java
```diff
@@ -88,6 +89,11 @@ else if (ex instanceof MismatchedInputException) {
 				serviceErrorCode = Status.BAD_REQUEST.getStatusCode();
 				serviceErrorListCode = serviceErrorCode;
 			}
+			else if (ex instanceof JsonParseException) {
```
csandersdev, line 92: I added a case for MismatchedInputException above which is a Jackson data bind exception I encountered in my testing. You have a block for JsonParseException. Both are parented by JsonProcessingException. I wonder if we should just handle both your case and mine in a JsonProcessingException block.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -83,26 +80,16 @@
  *
  */
 @Path(CohortEngineRestConstants.SERVICE_MAJOR_VERSION)
-@Api(value = "cohort")
-//Need this securityDefinition to get the "Authorize" button on the swagger UI to show up
-@SwaggerDefinition(securityDefinition = @SecurityDefinition(basicAuthDefinitions = { @BasicAuthDefinition(key = "BasicAuth") }),
-tags={@Tag(name = "FHIR Measures", description = "IBM Cohort Engine FHIR Measure Information"),
-						@Tag(name = "measures", description = "IBM Cohort Engine Measure Evaluation")})
+@Api()
```
csandersdev, line 83: It's a nitpick, but the empty parentheses here both me. Also, can you elaborate on why the value was removed? I thought that controlled the subpath of parts of the API. Are they recommending we don't use subpaths?

csandersdev (approved):
LGTM. I left a couple of minor points, but nothing I would block over.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
tabishop: The value was removed because it causes a "Cohort" Tag to get generated in the swagger.json. The tag causes an empty "Cohort" twistie section that doesn't have any API calls under it in the UI and the openApiValidator tool then complains about having a Tag that is not used.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortServiceExceptionMapper.java
tabishop: I'm not totally opposed to it, but just added the section for one I strictly encountered like you did. Not sure what other kind of exceptions that could get generated that would fall under the parent. I think if testing uncovers more, then we could change it.

PR #105, opened 2021-04-08
Update parameter documentation

msargentibm:
I noticed the docs still referenced the old, all-string version of parameters. I took a shot at updating the documentation to be more in line with the current parameter implementation.

Giving examples for each type was taking up a  bit of real estate, so I moved all of the parameter details and examples to a separate page. If folks would rather have those stay  on the getting started page I would be happy to move them back (or entertain an alternate refactoring if there are ideas).


csandersdev (approved):
LGTM. Thanks for cleaning up my mess.

PR #108, opened 2021-04-13
Exclude dependency for appscan vulnerability

tabishop:
Needed to exclude this dependency for an Appscan issue. Updating the version did not resolve the issue (resolution in the appscan issue was not worded clearly).


csandersdev (approved):
I started a side conversation with Tim about possibly using the Maven enforcer plugin to help manage some of these exclusions, but I've fine with what is presented here for now.

PR #109, opened 2021-04-13
Add unit test for "interval starts interval"

csandersdev:
It was noted that the CQL Author's Guide lists "interval begins
interval" as a valid operator, but it wasn't functioning in the engine.
After investigation, it was determined to be a documentation problem.
The correct operation is "interval starts interval". This unit test
validates that operator exists and is functional.

### DIFF
--- a/cohort-engine/src/test/java/com/ibm/cohort/engine/CqlEngineWrapperTest.java
+++ b/cohort-engine/src/test/java/com/ibm/cohort/engine/CqlEngineWrapperTest.java
@@ -615,4 +615,28 @@ public class CqlEngineWrapperTest extends BasePatientTest {
 			});
 		assertTrue( "Unexpected exception message: " + ex.getMessage(), ex.getMessage().contains("version and code system bindings are not supported at this time") );
 	}
+	
+	@Test
+	/**
+	 * This test exists to validate the the engine correctly evaluates CQL
+	 * that includes an "interval starts interval" expression. This was 
+	 * called out because the CQL Author's Guide documentation mentioned
+	 * "interval begins interval" as a supported feature and "begins"
+	 * isn't the correct operator name. 
+	 *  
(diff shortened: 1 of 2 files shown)


msargentibm (approved):

cohort-engine/src/test/java/com/ibm/cohort/engine/CqlEngineWrapperTest.java
```diff
@@ -615,4 +615,28 @@ private void runUnsupportedValueSetPropertyTest(String expression) throws ParseE
...
+		
+		final AtomicInteger resultCount = new AtomicInteger(0);
+		CqlEngineWrapper wrapper = setupTestFor(patient, "cql/temporal/IntervalStartsInterval-1.0.0.cql", "cql/temporal/FHIRHelpers.cql");
+		
+		wrapper.evaluate("IntervalStartsInterval", "1.0.0", /* parameters= */null, new HashSet<>(Arrays.asList("LHS Starts RHS")),
+				Arrays.asList(patient.getId()), (p, e, r) -> {
+					assertEquals(Boolean.TRUE, r);
+					resultCount.incrementAndGet();
+				});
+		assertEquals(1, resultCount.get());
```
deekim, line 640: I feel like this `assert` is redundant with the check on line `637`,
but not familiar enough with the code paths to object to your having it.

deekim (approved):

cohort-engine/src/test/java/com/ibm/cohort/engine/CqlEngineWrapperTest.java
csandersdev: It isn't completely redundant. It is possible that the CQL engine evaluated no matching defines and thus skipped the boolean check on line 637. This assert makes sure that code got executed.

PR #112, opened 2021-04-20
Swagger UI fix

tabishop:
The JsonProperty annotation was causing the swagger ui to improperly display the name of the parameter list in the results returned from the API call.


deekim (approved):

jaigoradia (approved):
Approved

PR #113, opened 2021-04-20
Patch the DataProvider to handle :in queries properly.

csandersdev:
There is an option on the TerminologyAwareDataProvider that allows it to
build FHIR search queries that use the token :in modifier to do
terminology expansion on the FHIR server during search. Use of this
feature has the potential to speed up performance by reducing round
trips to the FHIR server to get value set expansions and reducing round
trips to the search API for value sets with a large number of codes.

Under testing it was found that the OSS implementation that builds the
REST calls is broken. This appears to be an issue in the HAPI client
library. Prior to the use of the HAPI client to build queries, there was
a functional implementation in SearchParameterMap. That implementation
still exists and I added a code path that will fall back to it when any
modifier is found in the search parameters.

Since I was already patching the DataProvider, I added support for
configuring the pageSize requested from the FHIR server during search
requests. This will allow the user to override the very small default
values found in IBM and HAPI servers. Note that this is now configurable
on the server side in IBM FHIR 4.7.0. I set a default page size of 1000
because that is the maximum allowed in IBM FHIR.

I had to touch the CohortEngineRestHandler to add support for the new
configuration parameters, so I patched the documentation while I was in
there to provide a complete example of how to invoke the evaluation API.


cohort-cli/src/test/java/com/ibm/cohort/cli/MeasureCLITest.java
```diff
@@ -499,7 +499,7 @@ public void testZipFileKnowledgeArtifacts() throws Exception {
 		
 		Bundle emptyBundle = getBundle();
 		assertTrue( "Bundle should be empty", emptyBundle.isEmpty() );
-		mockFhirResourceRetrieval(get(urlMatching("/Condition.*")), emptyBundle);
+		mockFhirResourceRetrieval(get(urlMatching("/Condition.*&_count=500")), emptyBundle);
 		mockFhirResourceRetrieval(get(urlMatching("/Procedure.*")), emptyBundle);
```
deekim, line 503: I'm not familiar enough with this unit test to know if it would matter,
but should the other `Resource`s also have a `&_count=500` for consistency?

deekim (approved):

cohort-cli/src/test/java/com/ibm/cohort/cli/MeasureCLITest.java
csandersdev: The same _count parameter is used everywhere. I updated that one check just to make sure that it was being used with the expectation that the rest would fall in line. It is valid as-is, but we certainly could add the parameter to the other mocks.

csandersdev:
I've got a thread going with the OSS team in the cql_engine project [466](https://github.com/DBCG/cql_engine/issues/466) on whether or not the fallback solution here is valid. They moved away from their original implementation to solve a missing parameter problem. I'm going to leave this thread open until I get to a reasonable resolution point in that conversation.

csandersdev:
The OSS team says that the implementation was changed because parameter data was being dropped. The problem that was found will be addressed in HAPI FHIR 5.4 expected sometime in May. In the absence of more exhaustive unit tests, it is probably prudent to wait until the issue is fixed upstream.

cohort-cli/src/main/java/com/ibm/cohort/cli/CohortCLI.java
```diff
@@ -79,9 +79,15 @@
 		private LibraryFormat sourceFormat = DEFAULT_SOURCE_FORMAT;
 		
 		@Parameter(names = { "-i",
-		"--model-info" }, description = "Model info file used when translating CQL", required = false)
+				"--model-info" }, description = "Model info file used when translating CQL", required = false)
 		private File modelInfoFile;
-
+		
+		@Parameter(names = { "--expand-value-sets" }, arity = 1, description = "By default, ValueSet resources used in CQL are expanded by the terminology server. If the data server supports terminology and the token :in modifier, setting this flag to false will enable use of that functionality which should improve CQL engine throughput.", required = false )
+		private boolean expandValueSets = DEFAULT_EXPAND_VALUE_SETS;
```
dkwasny-ibm, line 86: Would it make more sense to name this `--disable-value-set-expansion` or something?  I try to steer away from having users put `true` or `false` after boolean parameters.  This is borderline pedantic though so if no one else has an issue, then I'm good.

docs/user-guide/getting-started.md
```diff
@@ -204,45 +222,57 @@ Usage: measure-engine [options]
...
       Possible Values: [ALL, BOOLEAN, NONE]
     --disable-retrieve-cache
       Disable the use of the retrieve cache.
+      Default: false
+    --expand-value-sets
+      By default, ValueSet resources used in CQL are expanded by the
+      terminology server. If the data server supports terminology and the
+      token :in modifier, setting this flag to false will enable use of that
+      functionality which should improve CQL engine throughput.
+      Default: true
```
dkwasny-ibm, line 241: Piggybacking on my earlier comment, you may want to add `Possible Values: [true, false]` to indicate to the user that they need to explicitly write `true` or `false` to use this parameter.

dkwasny-ibm (approved):
Looks good.  I have one pedantic thought regarding the new CLI parameters, but it can be ignored if desired.

PR #116, opened 2021-04-23
Wffhcohort 219 -- add parameters to measure reports

msargentibm:
Theses changes will add parameters and their values to the measure report generated by the engine. I used the `measure-parameter-value` format defined here: https://pages.github.ibm.com/watson-health-fhir-server/ig-common-data-model/StructureDefinition-quality-measure-report-patient.html.

`CDMMeasureEvaluation` contains the code to add the extensions to the MeasureReport. I was having trouble accessing all of the parameters in the Context object, so  I wound up passing in the parameter map and then building up a  list of parameter names using that as well as any parameters that are listed on the measure.

`MeasureReportParameterHelper` performs the conversion from CQL to FHIR objects and contains the logic to express date time objects in UTC.  I also tweaked the DatetimeParameter class for input parameters to default to UTC if a timezone is not provided. This PR does not contain changes to set the timezone for the overall engine execution, and it does not change the way parameters are handled if they are only specified on the FHIR Measure object.


cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -108,9 +113,10 @@ public CDMMeasureEvaluation(DataProvider provider, Interval measurementPeriod) {
 	 *                  measure evaluation
 	 * @param patientId Patient ID of the patient to evaluate
 	 * @param evidenceOptions MeasureEvidenceOptions to indicate whether or not to return evaluated resources and define level results
+     * @param parameterMap Map of parameter names to Parameter objects                       	   
 	 * @return MeasureReport with population components filled out.
 	 */
-	public MeasureReport evaluatePatientMeasure(Measure measure, Context context, String patientId, MeasureEvidenceOptions evidenceOptions) {
+	public MeasureReport evaluatePatientMeasure(Measure measure, Context context, String patientId, MeasureEvidenceOptions evidenceOptions, Map<String, Parameter> parameterMap) {
```
msargentibm, line 119: Passing in the parameters was done because I was unable to figure out how to access all of the parameters in the Context object without knowing the names of each parameter.  I wanted to cover the case where a parameter is passed in through Java, but not actually specified on the Measure.

I'm not sure if that case will happen in production, so I'd also be fine only  outputting parameters that were specified on the input Measure if folks think that would make more sense. In that case,  I would remove this change.

cohort-engine/src/test/java/com/ibm/cohort/engine/measure/MeasureEvaluatorTest.java
```diff
@@ -644,4 +657,52 @@ public void in_populations_supplemental_data() throws Exception {
...
+
+		Measure measure = getCohortMeasure("CohortMeasureName", library, INITIAL_POPULATION);
+		mockFhirResourceRetrieval(measure);
+
+		Map<String, Parameter> parameterMap = new HashMap<>();
+		parameterMap.put("integerParam", new IntegerParameter(1));
+		parameterMap.put("decimalParam", new DecimalParameter("1.0"));
+		parameterMap.put("stringParam", new StringParameter("1"));
+		parameterMap.put("booleanParam", new BooleanParameter(false));
+		parameterMap.put("datetimeParam", new DatetimeParameter("2020-01-01"));
```
csandersdev, line 680: It would be good to add a test for a DatetimeParameter that includes the timezone information to demonstrate and verify the behavior when the user provides a non-UTC timezone.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureReportParameterHelper.java
```diff
@@ -0,0 +1,107 @@
...
+import org.opencds.cqf.cql.engine.runtime.Ratio;
+import org.opencds.cqf.cql.engine.runtime.Time;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import ca.uhn.fhir.context.FhirVersionEnum;
+import ca.uhn.fhir.model.api.TemporalPrecisionEnum;
+
+
+public class MeasureReportParameterHelper {
```
csandersdev, line 28: Is this logic possibly more general purpose than just for MeasureReport parameters. It feels like this could be FhirTypeConverter in our namespace or something like that.

cohort-engine/src/test/java/com/ibm/cohort/engine/measure/MeasureReportParameterHelperTest.java
```diff
@@ -0,0 +1,295 @@
...
+		assertTrue(fhirTypeValue instanceof Range);
+		Range castResult = (Range) fhirTypeValue;
+		verifyBaseTypeAsQuantity(castResult.getLow(), startAmount, unit);
+		verifyBaseTypeAsQuantity(castResult.getHigh(), endAmount, unit);
+	}
+
+	/*
+	 *  Currently only supporting CQL intervals of type DateTime or Quantity.
+	 *  Support for intervals of other types may be added in the future.
+	 */
```
csandersdev, line 85: This comment feels like it is old/out of date. Is it still relevant?

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
csandersdev: Does the parameters map that you added as input contain the parameters that were defined on the Measure resource? I don't think it does. See MeasureEvaluationSeeder.create() for where the parameters defined on the measure report get copied into the Context.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -197,6 +205,56 @@ protected static void addDefineEvaluationToReport(MeasureReport report, CDMConte
...
+	
+	protected static Extension createParameterExtension(Context context, String parameterName) {
+		Object parameterValue = context.resolveParameterRef(null, parameterName);
+
+		Extension innerExtension = new Extension();
+		innerExtension.setUrl("http://ibm.com/fhir/cdm/StructureDefinition/parameter-value");
+		IBaseDatatype fhirParameterValue = MeasureReportParameterHelper.getFhirTypeValue(parameterValue);
+		// Do not create an extension for unsupported types
+		if (fhirParameterValue == null) {
+			return null;
```
csandersdev, line 242: Very nitpicky, but I would prefer the flow control here to be `if( fhirParameterValue != null )` and then have a single return at the bottom.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
msargentibm: It doesn't, but they are on the Measure itself and are retrieved later.

cohort-engine/src/test/java/com/ibm/cohort/engine/measure/MeasureReportParameterHelperTest.java
msargentibm: It should still be true.

cohort-engine/src/test/java/com/ibm/cohort/engine/measure/MeasureEvaluatorTest.java
msargentibm: I had considered this case, but it sort of felt like more of a test that was specific to DatetimeParameter. I'm fine still  adding a test to cover this.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/MeasureReportParameterHelper.java
msargentibm: I agree with it being potentially more general purpose, but I was nervous extending any  of the `FhirTypeConverter` classes in case they had some built-in calls happening behind the scenes that would make it tough control some of the output types (like DateTimes).

I went with wrapping the converter instead. If you would want to sit down and talk about what this might look like as a `FhirTypeConverter`, I'd be game.

cohort-engine/src/test/java/com/ibm/cohort/engine/measure/MeasureEvaluatorTest.java
msargentibm: Added a test.

cohort-engine/src/test/java/com/ibm/cohort/engine/measure/MeasureReportParameterHelperTest.java
csandersdev: I see it now. You have tests for other types, but they all test for null parameter output.

csandersdev (approved):
LGTM. All questions addressed to my satisfaction.

tabishop (approved):
Walked through with Matthew, LGTM

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/CDMMeasureEvaluation.java
```diff
@@ -197,6 +208,58 @@ protected static void addDefineEvaluationToReport(MeasureReport report, CDMConte
...
+			parameterNames.addAll(parameterMap.keySet());
+		}
+		
+		List<Extension> parameterExtensions = measure.getExtensionsByUrl(CDMConstants.MEASURE_PARAMETER_URL);
+		for (Extension e : parameterExtensions) {
+			ParameterDefinition parameterDefinition = (ParameterDefinition) e.getValue();
+			parameterNames.add(parameterDefinition.getName());
+		}
+		
+		parameterNames.forEach(x -> report.addExtension(createParameterExtension(context, x)));
```
dkwasny-ibm, line 234: Pedandic:  Would it make more sense to have this method return `List<Extension>`, and then add the new extensions to the report in the calling method?  I like to avoid side effecting when possible, but this isn't required in the slightest.

cohort-engine/src/main/java/com/ibm/cohort/engine/parameter/DatetimeParameter.java
```diff
@@ -22,6 +22,6 @@ public DatetimeParameter(String value) {
 	
 	@Override
 	public Object toCqlType() {
-		return new DateTime( getValue().replace("@",""), OffsetDateTime.now().getOffset() );
+		return new DateTime(getValue().replace("@",""), ZoneOffset.UTC);
```
dkwasny-ibm, line 25: Does this mean that all input datetimes will always be interpreted as UTC regardless of where the engine is running?  I think we went over this in one of our standup sessions, but it's all a blur at this point.

dkwasny-ibm (approved):
Seems OK.  Just some pedantic questions.

cohort-engine/src/main/java/com/ibm/cohort/engine/parameter/DatetimeParameter.java
msargentibm: The second argument is used for the timezone if no timezone exists in the String value used to create the DateTime.

So "2020-01-03T11:30:00+05:00" will be GMT+5, but "2020-01-03" will be UTC.

PR #118, opened 2021-04-26
New or Updated test files for testing REST API endpoints.

jaigoradia:
This PR consists of a mix of 26 new and/or modified test files as well as associated artifacts used by the automated test cases.  
The artifacts include: 
1. Four Measure zip files, one of which has a library deleted from it (missing dependencies scenario)
2. Two Value set spreadsheets and couple of text files for custom code system maps 
3. A Value set file in CSV format to represent invalid format for negative testing
4. A Measure file in json format used for negative testing. 

Other new files:
-----------------
A new file was created to represent the base request_data.json structure used by the Measure Evaluation REST API and stored in the config directory along with the local-ibm-fhir.json under tests/src/main/resources directory.

Four new yaml files were created one for each of the REST API endpoints as well as four new python driver files consisting of the actual automated tests to exercise each of the 4 REST API endpoints.  

A new utility TestClass.py file under the tests/src/main/python/rest/tests/utilities folder.

A new shell script was created under tests folder called setup-config-files-for-api-tests.sh  used to create variations of the request_data.json  and fhir config json files which are copied into the pod where the tests are run.  These files are specifically used for negative testing. 

Updated files:
---------------------------
Following automation framework files from before were updated:
1. run-fvttests.sh
2. setupEnvironmentVariables.sh 
3. create-taurus-pod.sh


jaigoradia:
One thing i would also like to mention here is that while trying to create tests for the Measure Evaluation API endpoint i had a hard time dealing with having evidenceOptions turned on.  There was just either too much json data being returned (could be multiple pages with ALL option and fairly substantial even with BOOLEAN option) and in case of includeEvaluatedResources i found the references kept changing on me every time the test was run so it was impossible to have a static expected results string. Bottom line i left "includeEvaluatedResources" : false and defineReturnOption: None for the tests.  I know the stuff works since i have done manual testing many many times now :).

csandersdev (approved):
There is a lot to digest here, but nothing especially jumped out at me as wrong. The only thing I might comment on is the number of sed calls being employed to update various files. That feels like it wants/needs a better solution. Something to investigate for the future.

PR #119, opened 2021-04-28
Wffhcohort 348 - Allow use of classpath FHIRHelpers

shelleyMurphy:
Added a method to read the classpath FHIRHelpers, adding it to the list of libraries provided.


cohort-engine/src/main/java/com/ibm/cohort/engine/FhirHelperBackupProvider.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Map;
+
+import org.hl7.elm.r1.VersionedIdentifier;
+
+public class FhirHelperBackupProvider {
+	private static VersionedIdentifier fhirHelpersVersion = new VersionedIdentifier().withId("FHIRHelpers");
+	private static final String FHIR_HELPER_DEFAULT_VERSION = "4.0.0";
+
+	FhirHelperBackupProvider(String fhirVersion){
+		fhirHelpersVersion.setVersion(fhirVersion);
```
dkwasny-ibm, line 22: A constructor modifying a static field feels pretty gnarly.  Maybe create a dedicated static method for setting the version instead?  That way you can make it `synchronized` if we ever need to.

You could also just create a `static` block like so.  This should also allow the use of `final`.
```
private static final Object THING;
static {
    THING = new Object();
    THING.setField("blah");
}
```

cohort-engine/src/main/java/com/ibm/cohort/engine/ClasspathLibrarySourceProvider.java
```diff
@@ -21,14 +21,13 @@ public ClasspathLibrarySourceProvider(List<String> libraryResources,
...
-				return new HashMap<LibraryFormat, InputStream>();
-			});
+			Map<LibraryFormat, InputStream> formats = sources.computeIfAbsent(vid, key -> new HashMap<>());
 			InputStream is = ClassLoader.getSystemResourceAsStream(resource);
 			if( is == null ) {
 				throw new IllegalArgumentException( resource );
 			}
 			formats.put(LibraryFormat.forString(resource), is);
 		}
+		FhirHelperBackupProvider.addClasspathFhirHelpers(sources);
```
dkwasny-ibm, line 31: Since it seems like all source providers want the correct `FHIRHelpers` instance loaded, would it make more sense to move this logic to the `MultiFormatLibrarySourceProvider` constructor?  That should allow us to move the logic from `FhirHelperBackupProvider` to `MultiFormatLibrarySourceProvider` as well.

We've already long gone down the path of super intelligent constructors, what's the harm in adding another in the base class?

cohort-engine/src/main/java/com/ibm/cohort/engine/FhirHelperBackupProvider.java
csandersdev: I agree with what Kwas is saying about changing static state in a constructor. That isn't a good idea. The singleton pattern would work here (e.g. store a single instance of FhirBackupHelpers as a static and expose it via a getInstance() method) though people frown on singletons these days because they make testing more difficult.

For my part, I am concerned with this design where we set a default FHIRHelpers version and then everybody gets that one version unless they fiddle with the version flag. We should let the CQL that is consuming FHIR Helpers determine what version it wants and then provide what they ask for. I get that the CQL Translator will only allow you to load one version of a library at a time and in those cases we should just throw an exception.

I think the hook point is MulitformatLibrarySourceProvider.getLibrarySource. We want that method to retrieve the FHIRHelpers from the classpath if it wasn't already loaded by the specific implementation.

The InJVMCqlTranslationProvider already uses the FhirLibrarySourceProvider and we are duplicating that work. We should consider whether or not that LibrarySourceProvider is still needed or not.

All of this is a good start, but I do want to point out that it doesn't address the Measure loading mechanism which is, unfortunately, a complete separate thing right now. There are a couple of touch points there included ResourceResolutionProvider for filesystem-based stuff and RestFhirLibraryResolutionProvider for the REST stuff.

csandersdev (changes requested):

msargentibm (approved):
Looks good. Thanks for confirming my understanding of the changes.

dkwasny-ibm (approved):

csandersdev (approved):
LGTM

PR #120, opened 2021-04-29
Update handling of measurement period

msargentibm:
I tried to bring `evaluatePatientMeasure(Measure measure, String patientId, String periodStart, String periodEnd, Map<String, Parameter> parameters, MeasureEvidenceOptions evidenceOptions)` in line with the evaluate measure spec: https://www.hl7.org/fhir/measure-operation-evaluate-measure.html. `periodStart` and `periodEnd` is now the only way to control the measurement period once we hit this method.

The main changes are:

- I added `MeasurementPeriodHelper` because the `DateHelper` class from the open  source engine was not properly resolving dates to the boundaries defined by the spec.
- I moved parameter overrides into the `MeasureEvaluationSeeder` class so that we could group together the different types  of parameter overrides and also to  highlight that "Measurment Period" is handled specially in order to respect the passed in `periodStart` and `periodEnd`.
- I also tweaked the MeasureReportBuilder to report the measurement period in UTC per previous discussions about how to handle output parameters.

I'm happy  to respond to feedback on here or talk through the code on a video  call if that would be helpful.

Edit:
I wanted to also add that we still support  overriding `Measurement  Period` with Java parameters given this update since that code is already a  part of the `DefaultMeasurmentPeriodStrategy` class.  These changes just assume that by the time we hit the final call to  `evaluatePatientMeasure`  that any overriding and  resolution to two dates has already been completed (whether by hand, or with a class implementing `MeasurementPeriodStrategy`).


cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
```diff
@@ -45,6 +44,8 @@
 
 	private boolean enableExpressionCaching;
 	private boolean debugMode = true;
+	
+	private static final String MEASUREMENT_PERIOD = "Measurement Period";
```
csandersdev, line 48: The CMS and HEDIS measures that I've looked at use this name for the measure period parameter, but it isn't defined anywhere in the specifications that I've seen. It makes me a bit nervous hard-coding it and then just expecting that the CQL author's will use it with that name. At one point I thought that Dan had made it configurable, but maybe that never happened. If we are forcing this specific name for the Measurement Period, I would like to see it at least documented in our user guide.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
```diff
@@ -98,27 +98,31 @@ public IMeasureEvaluationSeed create(Measure measure, String periodStart, String
...
 			String productLine) {
 
 		Context context = createDefaultContext(library);
 		context.registerLibraryLoader(libraryLoader);
 		context.registerTerminologyProvider(terminologyProvider);
 		context.registerDataProvider(modelUri, dataProvider);
 
-		context.setParameter(
-		null,
-                "Measurement Period",
```
csandersdev, line 118: Ok. Clearly hard-coded prior to your work. We could deal with this issue of the blessed parameter name in a separate ticket as needed.

csandersdev (approved):
LGTM

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
msargentibm: I think this was copied over from cqf-ruler: https://github.com/DBCG/cqf-ruler/blob/29e5a37fad5ec1890d74c0b087920049e76ef30d/r4/src/main/java/org/opencds/cqf/r4/evaluation/MeasureEvaluationSeed.java#L97

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/seed/MeasureEvaluationSeeder.java
```diff
@@ -98,27 +98,31 @@ public IMeasureEvaluationSeed create(Measure measure, String periodStart, String
 				.map(ParameterDefinition.class::cast)
 				.forEach(parameterDefinition -> setDefaultValue(context, parameterDefinition, dataProvider));
 
+		if (parameters != null) {
+			parameters.entrySet().stream()
+					.forEach(e -> context.setParameter(null, e.getKey(), e.getValue().toCqlType()));
+
+		}
+
+		// Set measurement perdiod last to make sure we respect perdiodStart
```
msargentibm, line 107:
```suggestion
		// Set measurement period last to make sure we respect periodStart
```

tabishop (approved):
LGTM

PR #123, opened 2021-05-04
Adjusted expected results for couple of tests to include stored parameters information being returned by evaluate measure rest api endpoint.

jaigoradia:


msargentibm:
Is there any risk to this breaking if parameters appear in a different order when this test is run multiple times?

jaigoradia:
> 
> 
> Is there any risk to this breaking if parameters appear in a different order when this test is run multiple times?

Does not appear to be any risk at this time. When i tried running the test multiple times via Swagger interface the parameters appeared to be returning in the same order every time.

msargentibm:
I phrased my first comment poorly.

What I meant to point out is that your test seems to only pass if the parameters appear  in the exact order you specify. However, the parameters on the measure report  are unordered according to  the CDM spec. If the underlying engine implementation changed the ordering, it looks like your test would fail. However, I think that the same parameters in a  different order would be a valid case.

As the measure reports the engine returns get more complicated, it may be worthwhile  to spend some time figuring out how to be a bit more precise in what the tests are validating and how that validation happens.

jaigoradia:
> 
> 
> I phrased my first comment poorly.
> 
> What I meant to point out is that your test seems to only pass if the parameters appear in the exact order you specify. However, the parameters on the measure report are unordered according to the CDM spec. If the underlying engine implementation changed the ordering, it looks like your test would fail. However, I think that the same parameters in a different order would be a valid case.
> 
> As the measure reports the engine returns get more complicated, it may be worthwhile to spend some time figuring out how to be a bit more precise in what the tests are validating and how that validation happens.

I get your point and will research how to improve the comparison of the measure report json objects in python for the future taking into consideration possibility of unordered parameters coming back for example.   I think for now we should just check in what i have updated. I don't for see any issue with these two tests breaking for how the code is currently working.  Can you go ahead and approve the PR for now.

msargentibm (approved):

PR #124, opened 2021-05-04
Wffhcohort 378 - Set timezone to UTC for engine. Make sure import parameters default to UTC

msargentibm:
This PR:
- Sets the default engine execution timezone to UTC, rather than the system timezone.
- Updates input measure parameter handling
    - Fixes handling of non-primitive parameters (Quantity, Period, Range, etc.)
    - Uses our cohort Parameter classes  as a common  point when converting to CQL

The timezone change is primarily in the CDMContext class. If  no timezone is provided when making the context, then UTC is used for the execution timezone. I made a  test CQL, `test-datetime-default-timezones.cql`, that is meant to exercise the engine's default timezone behavior  for date times in  CQL without timezones. It should pass as long as the default datetime timezone is UTC.

I also refactored the measure parameter handling in the MeasureEvaluationSeeder. It uses a new converter that takes FHIR parameters and converts them to our cohort Parameter classes (before eventually becoming  CQL). It's an extra step of conversion, but the thought was that this would allow us to have a common path to use during conversion. Also, various FHIR parameter types (Period, Range, Quantity, etc.) were not being converted to  CQL properly, so this new converter code handles those cases.

The remaining changes should generally be test cases to exercise the timezone and parameter changes. If folks have questions,  I'm more than happy to meet and go over any changes  in  depth.


cohort-engine/src/main/java/com/ibm/cohort/engine/measure/R4ParameterDefinitionWithDefaultToCohortParameterConverter.java
```diff
@@ -0,0 +1,149 @@
...
+public class R4ParameterDefinitionWithDefaultToCohortParameterConverter {
+
+	private R4ParameterDefinitionWithDefaultToCohortParameterConverter() {
+	}
+
+	public static Parameter toCohortParameter(ParameterDefinition parameterDefinition) {
+		Extension defaultValueExtension = parameterDefinition.getExtensionByUrl(CDMConstants.PARAMETER_DEFAULT_URL);
+
+		Parameter parameter = null;
+		if (defaultValueExtension != null) {
```
csandersdev, line 54: 90% of this method operates on the Extension and only this little bit at the top operates on the ParameterDefinition. Purely as a matter of code style, I would extract the part that deals with just the extension into a separate method ``toCohortParameter(Extension extension)`` and call it from here.

csandersdev (approved):
LGTM. I left a small comment on style, but I'll leave it up to you whether you want to make the suggested change.

cohort-engine/src/main/java/com/ibm/cohort/engine/measure/R4ParameterDefinitionWithDefaultToCohortParameterConverter.java
msargentibm: Would you want `toCohortParameter(Extension extension)` to be `public`, or are you fine leaving the primary public API requiring a  ParameterDefinition and having the extracted method stay `private`?

csandersdev: I would make it public. I have this idea in my head that we might want to convert and extension to a CQL parameter somewhere else and it might be useful.

shelleyMurphy (approved):

PR #125, opened 2021-05-04
Updating Jetty for vulnerability

jillrdoty:
PSIRT found.  I'm not going to report it since this is only used in test and therefore couldn't be exploited, but figured I'd update it anyway.


csandersdev (approved):
LGTM

msargentibm (approved):

PR #132, opened 2021-05-24
Test files updated to run in VPC environment

jaigoradia:
This PR consists of updates to tests files related to FVT tests as well as Performance tests so that they can run in the VPC environment against the FHIR server hosted in the VPC cluster and the knowledge tenant for measures/value sets instead of the default tenant as was the case in the CDT environment.


tests/src/main/resources/measureCLIExample-separate-measure-server.json
```diff
@@ -5,40 +5,40 @@
...
+            "response": "Evaluating: eb068c3f-3954-50c6-0c67-2b82d29865f0\nResult for Measure/wh-cohort-Colorectal-Cancer-Screening-Simple-1.0.0/_history/\nPopulation: initial-population = 1\n---\nEvaluating: 01268c61-8f89-6d50-4414-9e50d367804d\nResult for Measure/wh-cohort-Colorectal-Cancer-Screening-Simple-1.0.0/_history/\nPopulation: initial-population = 0\n---\nEvaluating: 6b6120d0-36b9-d2d5-639d-b77033cf5cee\nResult for Measure/wh-cohort-Colorectal-Cancer-Screening-Simple-1.0.0/_history/\nPopulation: initial-population = 1\n---\nEvaluating: 768202e4-b073-98e7-f3ab-2b7f467fea0b\nResult for Measure/wh-cohort-Colorectal-Cancer-Screening-Simple-1.0.0/_history/\nPopulation: initial-population = 0\n---",
```
msargentibm, line 8: I get that this change makes it so you don't have to update the version in the future. It also means that we no longer test for the change in https://jira.wh-sdlc.watson-health.ibm.com/browse/WFFHCOHORT-327 (unless you leave behind at least one test that cares about the report having the version in the report along with the measure id).

Just wanted to make sure that you were aware.

msargentibm (approved):
Seemed good overall. Just had a comment about a test case potentially disappearing (if that's something we care about).

tests/src/main/resources/measureCLIExample-separate-measure-server.json
jaigoradia: Since the measures got moved over to knowledge tenant from default tenant in earlier FHIR server we lost history '37'  so i was forced to not check for that while operating against knowledge tenant.  Yes i am aware about the version issue.

msargentibm: Mostly just wanted to make sure you didn't want at least one case left that checked for `Measure/wh-cohort-Colorectal-Cancer-Screening-Simple-1.0.0/_history/X` with an updated `X` for the knowledge tenant. It sounds like you're ok not testing for that anymore and that works for me.

PR #133, opened 2021-05-25
Updates to expected results for failing fvt rest api tests in new VPC environment

jaigoradia:
Updated expected results to fix 5 failing REST API tests post switchover to VPC environment.


csandersdev (approved):
LGTM

PR #136, opened 2021-06-09
R1 vpc updates

tabishop:
This PR contains cloud configuration and test file only changes to enable our release 1.0.0 code to run on a VPC environment. The 1.0.0 branch was cut before we did the switch to VPC so we needed to move these changes in. There are helm chart changes, terraform scripts needed to create the certificates (which was previously handled by a ticket to the CDT ops team) and test file changes.


jaigoradia (approved):
I not quite familiar with the terraform files but as far as test files go they look good to me. Approving with the assumption that the terraform files are correctly creating the certificates needed.

rickjstevens (approved):
LGTM

jillrdoty (approved):

PR #139, opened 2021-06-24
Release promotion updates for main

tabishop:
These are the same changes I put in the release branch to do the image promotion to the image repository and publish to the umbrella repo we would allow consumers to deploy from.


jaigoradia (approved):
Approved.

dkwasny-ibm (approved):

PR #140, opened 2021-06-25
Wffhcohort-419

shelleyMurphy:
Adding API for cohort evaluation for a set of patients. Originally this ticket was to be a single patient endpoint (WFFHCOHORT-419) but we refactored the design to be appropriate for patient sets as well (WFFHCOHORT-421).


cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -214,6 +235,122 @@ public CohortEngineRestHandler() {
...
+			@ApiResponse(code = 400, message = "Bad Request", response = ServiceErrorList.class),
+			@ApiResponse(code = 500, message = "Server Error", response = ServiceErrorList.class)
+	})
+	public Response evaluateCohort(@Context HttpServletRequest request,
+
+									@ApiParam(value = ServiceBaseConstants.MINOR_VERSION_DESCRIPTION, required = true, defaultValue = ServiceBuildConstants.DATE) @QueryParam(CohortEngineRestHandler.VERSION) String version,
+								   @ApiParam(value = CQL_DEFINE, required = true) @QueryParam(CohortEngineRestHandler.COHORT_DEFINE) String defineToRun,
+								   @ApiParam(value = PATIENT_IDS, required = true) @QueryParam(CohortEngineRestHandler.COHORT_PATIENT_ID) String patientIds,
+								   @ApiParam(value = CohortEngineRestHandler.COHORT_ENABLE_LOGGING, defaultValue = "false") @DefaultValue ("NA") @QueryParam(CohortEngineRestHandler.ENABLE_LOGGING) LoggingEnum enableLogging,
+								   @ApiParam(hidden = true, type="file", required=true) IMultipartBody multipartBody)
```
dkwasny-ibm, line 272: Pedantic: The above four lines seem to have spaces mingled with tabs.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -214,6 +235,122 @@ public CohortEngineRestHandler() {
...
+			@ApiResponse(code = 200, message = "Successful Operation"),
+			@ApiResponse(code = 400, message = "Bad Request", response = ServiceErrorList.class),
+			@ApiResponse(code = 500, message = "Server Error", response = ServiceErrorList.class)
+	})
+	public Response evaluateCohort(@Context HttpServletRequest request,
+
+									@ApiParam(value = ServiceBaseConstants.MINOR_VERSION_DESCRIPTION, required = true, defaultValue = ServiceBuildConstants.DATE) @QueryParam(CohortEngineRestHandler.VERSION) String version,
+								   @ApiParam(value = CQL_DEFINE, required = true) @QueryParam(CohortEngineRestHandler.COHORT_DEFINE) String defineToRun,
+								   @ApiParam(value = PATIENT_IDS, required = true) @QueryParam(CohortEngineRestHandler.COHORT_PATIENT_ID) String patientIds,
+								   @ApiParam(value = CohortEngineRestHandler.COHORT_ENABLE_LOGGING, defaultValue = "false") @DefaultValue ("NA") @QueryParam(CohortEngineRestHandler.ENABLE_LOGGING) LoggingEnum enableLogging,
```
dkwasny-ibm, line 271: I'm assuming the `defaultValue = "false"` is only used for documentation.  It should probably say "NA" as well right?

I also think `enableLogging` is probably a bad name for the parameter at this point.  Maybe use `loggingLevel` to line up with the CohortCLI?

cohort-engine/src/main/java/com/ibm/cohort/engine/LoggingEnum.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package com.ibm.cohort.engine;
+
+public enum LoggingEnum {
+	COVERAGE("coverage"),
+	TRACE("trace"),
+	NA(null);
+
+	private String debugLevel;
```
dkwasny-ibm, line 16: I'm a little confused on the purpose of the `debugLevel` variable.
All interactions with the enum directly refer to the uppercase enum value itself (including `getLoggingFromString()` below).
I also can't find any uses of `getDebugLevel()` being used in the code.

Unless it serves some deeper purpose I'm missing, I'd say remove the variable entirely.

cohort-engine/src/main/java/com/ibm/cohort/engine/BooleanEvaluationCallback.java
```diff
@@ -0,0 +1,49 @@
...
+	private static final Logger logger = LoggerFactory.getLogger(BooleanEvaluationCallback.class.getName());
+
+	@Override
+	public void onContextBegin(String contextId) {
+		logger.info("Context: " + contextId);
+	}
+
+	@Override
+	public void onContextComplete(String contextId) {
+		logger.info("---");
```
dkwasny-ibm, line 35: I think logging `contextId` on completion would be a good idea.
Since the server will be able to handle multiple concurrent evaluation requests, it would be nice to know which completion log line related to which `contextId`.

cohort-engine/src/main/java/com/ibm/cohort/engine/BooleanEvaluationCallback.java
```diff
@@ -0,0 +1,49 @@
...
+
+	@Override
+	public void onEvaluationComplete(String contextId, String expression, Object result) {
+		if( result != null ) {
+			if (result instanceof Boolean) {
+				if((Boolean) result){
+					passingPatients.add(contextId);
+				}
+				logger.info(String.format("Expression: \"%s\", Result: %s", expression, result));
+			}
```
dkwasny-ibm, line 46: Would it make sense to throw an error if a boolean is not returned here?
A user accidentally submitting a CQL define that doesn't return a boolean value seems very possible, and could cause confusion when the result mimics if a patient simply wasn't in the measure (aside from the missing log statement).
I'm leaning towards failing hard and fast here, but I'd like to hear any other opinions on this as well.

I'm not sure if we want to also do something similar for `null`.
If we want `null` to be a legit value, then are we OK with it being functionally equivalent to `false` (aside from the missing log statement)?
My gut says to fail hard and fast here as well until we identify a real usecase where `null` is a sensible return value.

cohort-engine/src/main/java/com/ibm/cohort/engine/LoggingEnum.java
```diff
@@ -0,0 +1,35 @@
...
+
+	public static LoggingEnum getLoggingFromString(String value){
+
+		try{
+			return LoggingEnum.valueOf(value);
+		}
+		catch(Exception e) {
+			return LoggingEnum.NA;
+		}
+	}
```
dkwasny-ibm, line 30: Pedantic: I don't see this method being used anywhere.  I'm OK with keeping it around if we think people will want to do a "parse value with default" operation in the future.

dkwasny-ibm (changes requested):
Mostly small changes and questions.
The functionality overall seems fine.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
shelleyMurphy: fixed
\

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
shelleyMurphy: fixed

cohort-engine/src/main/java/com/ibm/cohort/engine/LoggingEnum.java
shelleyMurphy: Yeah, I thought I needed it, got distracted, and then never refactored the class. Whoops >.< but this and your other comment re the enum are both fixed

cohort-engine/src/main/java/com/ibm/cohort/engine/BooleanEvaluationCallback.java
shelleyMurphy: done

cohort-engine/src/main/java/com/ibm/cohort/engine/BooleanEvaluationCallback.java
shelleyMurphy: I think I had that at some point, but cut it out when shuffling things around. I agree that it's a good idea to fail loudly, and have added that

cohort-engine/src/main/java/com/ibm/cohort/engine/BooleanEvaluationCallback.java
```diff
@@ -32,6 +32,7 @@ public void onContextBegin(String contextId) {
 
 	@Override
 	public void onContextComplete(String contextId) {
+		logger.info(contextId + "complete");
```
dkwasny-ibm, line 35: Adding a space before `complete` will make the log line a little more readable

cohort-engine/src/main/java/com/ibm/cohort/engine/BooleanEvaluationCallback.java
```diff
@@ -44,6 +45,11 @@ public void onEvaluationComplete(String contextId, String expression, Object res
 				}
 				logger.info(String.format("Expression: \"%s\", Result: %s", expression, result));
 			}
+			else {
+				throw new RuntimeException("Only boolean CQLs are currently supported!");
+			}
+		} else {
+			throw new RuntimeException("Null result is unsupported");
```
dkwasny-ibm, line 52: I think these two exceptions should log the parameters like what happens on line 46.
In fact, I think it's probably a good idea to extend the log statement on line 46 to include the context id.
I guess logging `result` isn't necessary on the null case specifically.

dkwasny-ibm (changes requested):
Some small log requests.

cohort-engine/src/main/java/com/ibm/cohort/engine/BooleanEvaluationCallback.java
shelleyMurphy: Fixed

cohort-engine/src/main/java/com/ibm/cohort/engine/BooleanEvaluationCallback.java
shelleyMurphy: Fixed

dkwasny-ibm (approved):

jaigoradia (approved):
Approved.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -214,6 +235,122 @@ public CohortEngineRestHandler() {
...
+			List<String> patientsToRun = Arrays.asList(patients);
+
+			evaluator.setLibraryLoader(new TranslatingLibraryLoader(provider, translationProvider));
+
+			VersionedIdentifier versionedIdentifier = new DefaultFilenameToVersionedIdentifierStrategy().filenameToVersionedIdentifier(cqlAttachment.getDataHandler().getName());
+
+			//todo is it fair to assume the default strategy will be appropriate? Currently it would be being used elsewhere
+			//for right now we're not going to support custom parameters, since I think that would be instead of uploaded cql and define
+			BooleanEvaluationCallback callback = new BooleanEvaluationCallback();
+			//todo here
```
msargentibm, line 335: Are the todo comments in this file there on purpose still?

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -214,6 +235,125 @@ public CohortEngineRestHandler() {
...
+							, value = CohortEngineRestConstants.DARK_LAUNCHED_COHORT_EVALUATION)
+			})
+	}
+	)
+	@ApiImplicitParams({
+			// This is necessary for the dark launch feature
+			@ApiImplicitParam(access = DarkFeatureSwaggerFilter.DARK_FEATURE_CONTROLLED, paramType = "header", dataType = "string"),
+			// These are necessary to create a proper view of the request body that is all wrapped up in the Liberty IMultipartBody parameter
+			@ApiImplicitParam(name=REQUEST_DATA_PART, value=EXAMPLE_REQUEST_DATA_JSON, dataTypeClass = MeasureEvaluation.class, required=true, paramType="form", type="file"),
+			@ApiImplicitParam(name=CQL_PART, dataTypeClass = File.class, required=true, paramType="form", type="file" )
```
csandersdev, line 259: The docs here don't line up with what is used inside the method. The name and dataTypeClass are wrong in the first parameter and the name for the second parameter should be  `CQL_DEFINITION`.

csandersdev (commented):
It would be good to add a test case for CQL that fails translation. I had to do some work in my IBM FHIR impl to pretty that up for the end user. I expect we will have to do something similar here.

cohort-engine/src/main/java/com/ibm/cohort/engine/CqlEvaluator.java
```diff
@@ -453,11 +470,13 @@ protected void evaluateWithEngineWrapper(String libraryName, String libraryVersi
...
 	 *                       specified <code>expressions</code> will be executed. At
 	 *                       least one value is required.
+	 * @param enableLogging  The level of logging enabled, either TRACE, COVERAGE, or NA
+	 *
 	 * @param callback       callback function for receiving engine execution events
 	 */
 	public void evaluate(String libraryName, String libraryVersion, Map<String, Parameter> parameters,
-			Set<String> expressions, List<String> contextIds, EvaluationResultCallback callback) {
-		evaluateWithEngineWrapper(libraryName, libraryVersion, parameters, expressions, contextIds, callback);
+			Set<String> expressions, List<String> contextIds, LoggingEnum enableLogging, EvaluationResultCallback callback) {
```
msargentibm, line 478: Copy pasting  a comment I left on Dan's PR for a similar change he made:

>I'll bring this up as a topic at a team meeting, but do we need to be mindful of breaking public apis since we have had a 1.0.0 release at this point? Should we aim for backwards compatibility so we don't potentially break consumer code?

msargentibm: Talked with team. Don't worry about this change.

PR #141, opened 2021-06-25
Allow non-semantic version strings for resources

msargentibm:
Change logic in ResourceResolutionProvider.java to index resources no matter the version string (previously threw an error if version was not a semantic version).

When retrieving resources and not providing a version to use during the retrieval, only consider resources with semantic versions when calculating the latest version. When a version is provided, attempt an exact match using the provided version.


csandersdev (approved):
Discussed with Matt and LGTM.

jaigoradia (approved):
LGTM

shelleyMurphy (approved):

PR #143, opened 2021-07-06
wffhcohort407 -- Add ability to cache FHIR types with the R4FhirModelResolver

msargentibm:
Edit: Adding new description. Leaving the old  description below.
Edit2: Putting use of JCache on  hold for now.

# New Description
The open source project had model resolver caching added in the `CachingModelResolverDecorator` class in the `cql-evaluator/engine.evaluator` project. After a discussion in the morning meeting, we decided to copy, paste, and modify the class rather than including an entirely new dependency.

~The main modifications I made to  the class were to replace the Map-only approach in the open source code with one that uses JCache Caches. I kept the pattern of keeping caches per package name to minimize my overall modifications to the open source code, but would be happy to revisit this since we only really support R4 FHIR at the moment.~

~I would also appreciate an extra close review/sanity check of the patterns I use for initializing the caches. It feels like this is the sort of change where a little mistake could easily slip by, and I'd like to catch any oddities early if possible.~

Thanks!

# Original description
I added  the ability to cache  some of the FHIR model resolution  results by  adding the `CachingR4FhirModelResolver` class. It wraps the `R4FhirModelResolver` class and adds a shared static cache to store results for the `resolveType()` method. I am using  the default cache configuration  settings from JCache, rather than making this cache fully configurable.

I addition to the new class, I changed our code in various places to use the new caching resolver. Most of those changes were just replacing `R4FhirModelResolver` with `CachingR4FhirModelResolver`.
When it came to the `R4DataProviderFactory`, I added a new `createDataProviderMap` method rather than changing the existing method so that our public API was preserved (I wanted to be backwards compatible with our public API as of the `1.0.0` release).

The ultimate goal is to get a FHIR model caching solution accepted back in the open source code base (I have an issue opened up with them to start a conversation about what a change in their code might look like). At that point, the changes in this PR could likely be reverted. As such, I did not focus on making this caching layer fully customizable. If folks think we might want more knobs to turn for this work, I am open to suggestions.


msargentibm:
Going to copy the pattern in https://github.com/DBCG/cql-evaluator/blob/master/evaluator.engine/src/main/java/org/opencds/cqf/cql/evaluator/engine/model/CachingModelResolverDecorator.java, but include JCache instead of using maps.

cohort-engine/src/main/java/com/ibm/cohort/engine/r4/cache/CachingModelResolverDecorator.java
```diff
@@ -0,0 +1,170 @@
...
+public class CachingModelResolverDecorator implements ModelResolver {
+	private static final String CACHE_PREFIX = "model-resolver";
+	
+	private static final String PER_PACKAGE_CONTEXT_RESOLUTION_CACHE_PREFIX = CACHE_PREFIX + "-context-resolution";
+	private static final String PER_PACKAGE_TYPE_RESOLUTION_BY_NAME_CACHE_PREFIX = CACHE_PREFIX + "-type-resolution-by-name";
+	private static final String PER_PACKAGE_TYPE_RESOLUTION_BY_CLASS_CACHE_PREFIX = CACHE_PREFIX + "-type-resolution-by-class";
+	
+	static final Map<String, Map<String, Cache<String, Object>>> perPackageContextResolutions = new HashMap<>();
+	static final Map<String, Cache<String,Class<?>>> perPackageTypeResolutionsByTypeName = new HashMap<>();
+	static final Map<String, Cache<Class<?>,Class<?>>> perPackageTypeResolutionsByClass = new HashMap<>();
```
csandersdev, line 25: I find this pattern where we are keeping a Map (or Map of Map) of JCache cache objects strange.  I can see the need to keep separation of resolutions by package name, but the hybrid management of JVM managed and also cached managed data feels wrong to me. It lessens or defeats the purpose of having the JCache objects to begin with. If we are using JCache, I would expect that the top level objects be JCache cache objects. Ideally, the cache would be done with a cache key object that contains all the necessary data so that the cache structure is just a flat name-value-pair structure. In the case of these methods, the CacheKey could be a Pair or Triple object with the appropriate generic parameters.

I know we have a similar pattern in our other cache work, but the creation of these caches dynamically in the consuming classes also seems wrong to me. In my experience, initialization for the caches should be done up front during application startup. For something like ehcache, there was an XML configuration file where you defined the caches you would use in your application. Those were created when you initialized the caching subsystem from that XML file. Caffeine seems to have a similar thing using [lightbend formatted config files](https://github.com/ben-manes/caffeine/wiki/JCache), but it is very poorly documented. Everything wants us to use Spring bean management to enable the caching subsystem. I'd very much like to do that, but it is much bigger than the current ticket.

I am a big proponent of full caching subsystems, but as I've been staring at this I don't think most of those features would be beneficial here. I'd be willing to revisit the JCache mandate and decide to just use the copy/pasted Map-based implementation if you want. We would miss out on weak references, but other features like a configurable eviction policy and potential to adapt to a distributed cache model don't seem like they'd ever be important here. The data in the caches should be relatively small and reach a steady state after a while.

msargentibm: Thanks for the review. Your feedback makes sense and some variations of these thoughts crossed my mind as I was working on this code.

> the CacheKey could be a Pair or Triple object with the appropriate generic parameters.

Agree on the map bit being wonky. I don't know why having a proper key for each cache didn't cross my mind (even some of the retrieve cache work uses a proper key).

> initialization for the caches should be done up front during application startup.

I thought about this as well. I didn't yet spend the time to learn how to configure cache creation at startup, but could do that.

> I am a big proponent of full caching subsystems, but as I've been staring at this I don't think most of those features would be beneficial here. I'd be willing to revisit the JCache mandate and decide to just use the copy/pasted Map-based implementation if you want.

I'll at least try to incorporate the other feedback you had and see what folks think of a "cleaned up" JCache implementation. If that's still overkill, I'd be happy to use the copy pasted class.

msargentibm: > verything wants us to use Spring bean management to enable the caching subsystem. I'd very much like to do that, but it is much bigger than the current ticket.

I read up a bit more on this after the morning  meeting, and agree that it seems bigger than this ticket how to set all of this up properly. I'll pivot to using the copy-pasted `Map` for now. I can add a backlog ticket/spike to assess the need for JCache here (or perhaps a spike for cache creation at application startup across the board).

msargentibm:
Putting JCache on hold for the moment and am using the copy-pasted `Map` caching from the open source code. Will open a ticket for follow up JCache work or a JCache spike.

csandersdev (approved):
LGTM

shelleyMurphy (approved):
lgtm

PR #144, opened 2021-07-07
Enhance the logging around library loading and translation

csandersdev:
While doing some debugging with QA on an issue related to CQL
translation, we found that the detail provided to clients in our CQL
Translation errors was inadequate to isolate the problem. The changes in
this PR help narrow in on the problem. Specifically, it adds to
translation errors the library name, library version, and start and end
positions of the expression that caused the failure. Logging was also
added for translation warnings and library loading.


cohort-engine/src/main/java/com/ibm/cohort/engine/translation/InJVMCqlTranslationProvider.java
```diff
@@ -132,4 +136,24 @@ public ModelInfo convertToModelInfo(Reader modelInfoReader) {
...
+	
+    /**
+     * Some of this was adapted from the CQL Translation Server TranslationFailureException.
+     * 
+     * @param translationErrs List of translation errors.
+     * @return String representation of the list of translation errors.
+     */
+    private static String formatMsg(List<CqlTranslatorException> translationErrs) {
+        StringBuilder msg = new StringBuilder();
+        msg.append("Translation failed due to errors:");
```
msargentibm, line 148: Does translation actually fail in the `warning` case?

msargentibm (approved):

shelleyMurphy (approved):

PR #145, opened 2021-07-09
Wffhcohort 435 - Search parameter caching

msargentibm:
Work in progress for search parameter caching.

I copy pasted methods and overwrote some of them. I'd like a better way to handle the FhirContext in `CachingSearchParameterResolver`,  but didn't have any great ideas initially since the context is private in the parent class.


csandersdev (commented):
I notice we are using JCache here after we recently rolled back that decision in the ModelResolver. There was a lot more complexity in ModelResolver, so I get it. I just wonder about consistency. Would this be better off as a basic Map until we figure out the Spring stuff?

I would love to see a comment on each of the methods that basically says "copy/pasted from CQL Engine 1.5.1. Revisit on upgrade of that dependency" or something along those lines.

msargentibm:
Closing for now. Mentioned last week in standup that I am putting wffhcohort-435 on hold for the time being since the open source folks are still debating how they want to handle search parameter caching.

PR #148, opened 2021-07-13
apply expression cache clear patch from cql-ruler

deekim:
ports over the changes in [DBCG/cqf-ruler#154](https://github.com/DBCG/cqf-ruler/pull/154/files#diff-2ba2e6fa2865691c2557d753dd53f08a88e3294d3cb84f70f6a5f66119efc8f3)

not exactly sure this is the best way to handle this,
but with so much code being copied over, this was the least intrusive change.

---

as Corey mentioned, the ability to use `setAccessible` won't work going into java 11,
so I also opened up a PR in https://github.com/DBCG/cql_engine/pull/494 to expose that functionality.

---

## todo
- [x] add unit test case


shelleyMurphy (approved):

csandersdev (commented):
I wondered if we would have the same problem in the CqlEvaluator, but it appears that we are creating the CQL Context object on a per-patient basis, so it won't have this problem.

I checked out the branch to my laptop and ran the MeasureEvaluator unit tests on JDK11 and they passed ok. This leads me to believe that targeting Java 8 during compilation will let this run on newer VMs despite it violating the modern security constraints.

It would be nice to have a unit test that confirms the behavior is working as expected.

PR #153, opened 2021-08-01
New swagger code autogenerated to include API interfaces for the two new REST endpoints.

jaigoradia:
This PR is to merge newer version of swagger code generated to include the API interfaces for the two new REST endpoints.  There is no need to spend time reviewing this PR since the code is generated via a tool.


msargentibm (approved):
Rubber stamp approval

PR #154, opened 2021-08-02
New and updated test files for testing the two new REST API endpoints

jaigoradia:
This PR includes the test artifacts necessary to test the two new REST API endpoints: 
a. cohort evaluation
b. measure evaluation using a list of patients.

The files:
1. cohort-evaluation-api-request_data.json and
2. evaluate-patient-list-measure-api-request_data.json   represent the the request_data.json objects passed in as parameters to the two endpoints. These were created based on the documentation provided on the swagger interface.  Specific place holders are used at strategic places in the json checked into GIT, eventually replaced with real values via sed operations that are done in the script called setup-config-files-for-api-tests.  

**Multiple copies of the request_data files eventually get created tailored for the specific test case being executed.  
Note: These files are only created at runtime but then they go away with the test pod that end since all test artifacts are copied into the test pod that is spun off in the framework. 

Both happy path and negative tests are run for testing the endpoints.  All patients are retrieved from test-fvt tenant and all measure and value set artifacts are retrieved from the knowledge tenant.


msargentibm (approved):
Tried my best to look through everything, but I'll admit that the tests all tend to blur together after a bit.

PR #156, opened 2021-08-03
explicitly specify provider versions to avoid being broken when a newer

tabishop:
version is published that doesn't work with the backlevel version of
terraform the toolchains are using


msargentibm (approved):

jaigoradia (approved):
Approved.

PR #157, opened 2021-08-05
Wffhcohort 525

shelleyMurphy:
Updating documentation to hopefully make our endpoints arguments clearer and to extend our code

eta, if you see anything specific you'd like changed, formatting or informationally, feel free to either comment or update the branch directly, I know I overly-gravitate to bulleted lists


docs/user-guide/rest_endpoints.md
```diff
@@ -0,0 +1,34 @@
...
+- measureId: The id of the measure to run. You do not need a measureId and an identifier, you only need one or the other.
+- identifier: The identifier of the measure to run. If you provide this, you do not need a measureId but you do need a version.
+- version: The version of the measure to run. This is only necessary if you're locating your measure with the identifier rather than the measureId.
+- parameters: A place to specify any neccessary input parameters for your define. For example, if your define requires a measurement period this is where you would supply it.
+- evidenceOptions: Two options for adding extra information to the returned measure report.
+	- includeEvaluatedResources: Whether or not to include the resources that were evaluated to determine if patients succeed or fail in the measure report.
+	- defineReturnOption: If this is set to ALL, will return add the define result to the report regaredless of the type. If set to BOOLEAN and the define returns a boolean type, the boolean will be added to the measure report. If it is set to BOOLEAN and the define return type is not BOOLEAN, nothing will be added to the measure report.
```
csandersdev, line 13: When we discussed as a team, I believe the agreement was that we weren't going to duplicate the documentation that is already stated in the Swagger in the user guide. The intention was to provide additional, detailed behavior on API behavior that wasn't REST specific. Mainly, I had an interest in creating a section (either in the MeasureEvaluator section of the getting started guide or a completely new page) that detailed the evidence options because those are relevant whether you are a Java API or REST API consumer. I really think that everything related to quality measure evaluation could/should be a separate page in the user guide.

For "includeEvaluatedResources":
1) I'd like it to be clear about _where_ the evaluated resources are included and how. Specifically, the MeasureReport resource's [evaluatedResources ](https://www.hl7.org/fhir/measurereport-definitions.html#MeasureReport.evaluatedResource) property is populated when this flag is set to true. It is provided as an option because the data for that flag can get quite large and may not be necessary for the consuming application. We can drive storage and processing efficiencies by leaving them out.
2) Need to be clear about what the default value is here.

For "defineReturnOption": 
1) you've got a grammatical error near "return add". Probably choose one or the other verb. 
2) I think we should expand this section to be absolutely clear about how the data is rendered in the MeasureReport. I would mention the CDM IG extension by URI, show an example, and specifically call out that define statements that return whole resources are rendered as references to the resource and not the resource itself.
3) I think we should call out that this option is intended for users that need to do debugging of the CQL that is executed in support of the measure evaluation and that it defaults to NONE.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
```diff
@@ -236,6 +238,7 @@
 			"<p>The <code>patientIds</code> is a comma separated list of patients to run. Supplying a single patient does not need any trailing commas.</p>" +
 			"<p>The parameter types and formats are described in detail in the <a href=\"http://alvearie.io/quality-measure-and-cohort-service/#/user-guide/parameter-formats?id=parameter-formats\">user guide</a>.</p>" +
 			"<p>The <code>loggingLevel</code> will determine how much and what type of logging to provide. The options are NA, COVERAGE, and TRACE.</p>" +
+			"<p>For more detailed information about the cohort endpoint, check out our <a href=\"http://alvearie.io/quality-measure-and-cohort-service/#/user-guide/rest_endpoints?id=cohort-endpoint\">documentation</a> here.</p>" +
```
jaigoradia, line 241: @shelleyMurphy : As i had pointed it out to you the other day can you add a statement or two regarding the fact that in the Example contents for the request_data.json you are showing for the cohort evaluation endpoint the measurement period interval start type and end type are different. One is date and other is datetime.  Need to clarify for users that either one can be used but the start and end should match each other for any given instance of use. You cannot have one as date type and the other as datetime type.

jaigoradia (changes requested):
Left comments.

docs/user-guide/rest_endpoints.md
shelleyMurphy: Ok, I had gotten exactly the opposite out of that conversation, but I can move these and delete this page.

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
shelleyMurphy: I'll just make them the same value and hope that people look at the parameter type/format url

docs/user-guide/rest_endpoints.md
```diff
@@ -0,0 +1,34 @@
...
+## Parameters
+- entrypoint: The specific cql file that contains the define you plan to run. If you change the entrypoint to a different cql file, you will also need to ensure your defineToRun is defined in that file.
+- defineToRun: This is the specific define within the specified cql that will be evaluated. Altering this parameter will change which define is specifically evaluated. NOTE: Currently, only boolean defines are supported.
+- patientIds: A comma-separated list of patients that you intend to run. A single patient does not require trailing commas.
+- loggingLevel: A flag for setting logging levels. Default is NA (no additional logs), or you may specify TRACE for additional logs.
+- parameters: A place to specify any neccessary input parameters for your define. For example, if your define requires a measurement period this is where you would supply it.
+- cql_definition: A zip file containing all the necessary cql files for 
+## Output
+Given a list of patients and a define, the return value will be the list of patients who passed the define, aka those patients who, when evaluated against the define, returned true.
```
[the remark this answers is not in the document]
shelleyMurphy: This file has been deleted per Corey's comment

cohort-engine-api-web/src/main/java/com/ibm/cohort/engine/api/service/CohortEngineRestHandler.java
jaigoradia, line 241: I see you have links point to sections internally in user guide for rest endpoints in each of the EXAMPLE REQUEST_DATA JSON Strings.  If you removed anything REST specific from user guide those should probably not be there either??

jaigoradia (changes requested):
Added one more comment regarding REQUEST_DATA_JSON strings having links back to user guide for rest endpoints for the 3 endpoints.  Should those be removed?

jaigoradia (approved):
Approved.

PR #161, opened 2021-08-19
Wffhcohort 465 terraform for spark on k8s

tabishop:
Terraform scripts for spark on k8s cluster configuration and a fix to a test case that was needed due to a change to the error message returned by an updated version of an OSS javax.validation package we included.


terraform/main.tf
```diff
@@ -1,70 +1,135 @@
...
-    name = var.namespace
-  }
+##############################################################################
+# Kubernetes signed certificate creation for REST API server
+##############################################################################
+
+module "tls_cert" {
+  source                    = "./modules/tls_cert"
+  namespace                 = var.tsl_cert_namespace
+  service_names             = var.tsl_cert_service_names
```
jaigoradia, line 78: Are these variables suppose to de-reference var.tls_* or var.tsl_*?  ie. Is that a typo ?

jaigoradia (approved):
Approve based on my very limited knowledge of terraform. Just eyeballed the changes for obvious errors.  Commented at one place since there looked like a typo although i could be wrong.

terraform/main.tf
tabishop: Good catch, fixed.

PR #167, opened 2021-09-08
Bug fixes and performance improvements

csandersdev:
While testing under heavy load, several issues with the Spark application were identified. The CQL libraries and CQL jobs specification files were being reread from COS for every evaluation. Additionally, it was later found that each CQL context object that was created caused the application to reread an XML file for initialization of the UcumEssenceService and that was taking an inordinate amount of processing cycles. Caching was added for all of these parts of the program.

Simultaneous with the performance testing several bugs were found in the application. First, the CQL jobs specification was not being filtered properly per job evaluation, so that all jobs were being evaluated for all contexts. Secondly, the merge logic was broken because the merged dataset was not captured and sent downstream for processing. Those bugs were fixed in the branch.

Lastly, some missing features were added. This includes the ability using Spark SQL metadata to define a default system for a code column (vs. retrieving the code value from a separate column) and allowing the application to load CQL libraries from a classpath source and using a priority based search mechanism. Also, Spark accumulators were added to track the number of context executions processed so far by an executor and the number of context aggregations processed.

The branch code was tested against the s3://sargent-mock-test-data/input data (roughly 5 million patient records with data in multiple tables) and processing completed across 2 executors in just under 5 hours.


cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -98,16 +114,24 @@
...
 
     @Parameter(names = { "--debug" }, description = "Enables CQL debug logging")
     public boolean debug = false;
 
-    public SparkTypeConverter typeConverter;
+    protected SparkTypeConverter typeConverter;
+
+    protected static ThreadLocal<CqlEvaluationRequests> jobSpecification = new ThreadLocal<>();
+    protected static ThreadLocal<CqlLibraryProvider> libraryProvider = new ThreadLocal<>();
+    protected static ThreadLocal<CqlTerminologyProvider> terminologyProvider = new ThreadLocal<>();
```
dkwasny-ibm, line 134: Pedantic: I'm not sure what to say, but I feel like some comment stating that these providers need to be unique per thread would be helpful for future devs.  Making any change to their scope will conflict with how `CQLContextFactory` is currently designed (namely, the `CONTEXT_CACHE` map).

cohort-evaluator/src/main/java/com/ibm/cohort/cql/evaluation/CqlContextFactory.java
```diff
@@ -26,70 +32,206 @@
...
-    /**
-     * Controls whether CQL engine trace logging is enabled. Defaults to false.
-     */
-    private CqlDebug debug = CqlDebug.NONE;
+    protected static class ContextCacheKey {
+        public CqlLibraryProvider libraryProvider;
+        public CqlTerminologyProvider terminologyProvider;
+        public CqlLibraryDescriptor topLevelLibrary;
+        public ZonedDateTime evaluationDateTime;
+        public Map<String,Object> parameters;
```
dkwasny-ibm, line 45: Pedantic: Normally, I'd shy away from throwing `final` suggestions your way, but I think in this case it's extra important that these fields do not change.
Since you already have a full constructor setup, it should be easy to set these as `public final` without issue.

dkwasny-ibm (approved):
Some relatively small comments.
Nothing worth blocking the PR over.

PR #169, opened 2021-09-16
Update spark-history-server to use latest spark-cos-secret contents

csandersdev:
I updated the Spark History Server deployment to use values from the newly formatted `spark-cos-secret` that is based on the IBM S3 FS plugin. We don't have endpoint ID as part of that secret, so it needed to come from somewhere else. I set a default value of the US geography endpoint with the assumption that the CD deployment pipeline will override to set whatever is truly used in the deployment environment. As part of that, I updated the overrides files in the cohorting-config repo. This was tested in cicdtest and works.


jaigoradia (approved):
LGTM

dkwasny-ibm (approved):

PR #170, opened 2021-09-17
Filtering fix

msargentibm:
Filtering logic was happening, but the resulting object wasn't being passed through to the final  evaluate. I moved the logic to the final evaluate statement.

I'm not a huge fan of the extra method I added, but it made it easy to write tests quickly.

There was also a NullPointerException for the CodeKey class if System was null and equals was called. I added a test and a code fix.

### DIFF
--- a/cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
+++ b/cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
@@ -3,6 +3,7 @@ package com.ibm.cohort.cql.spark;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 import java.io.File;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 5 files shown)


csandersdev (approved):
LGTM. Thanks for doing this.

msargentibm:
I should note that I've only briefly inspected the output from a test run to make sure that measures were being reported in their expected contexts. I didn't do a deep dive into any individual results.

jaigoradia (approved):
Approved

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -136,6 +139,107 @@ public void testAllTypesEvaluationSuccess() throws Exception {
...
+                makeEvaluationRequest("A"),
+                makeEvaluationRequest("A"),
+                makeEvaluationRequest("B")
+        );
+
+        requests.setEvaluations(evaluations);
+
+        List<CqlEvaluationRequest> cqlEvaluationRequests = evaluator.filterRequests(requests, "A");
+
+        assertEquals(2, cqlEvaluationRequests.size());
```
dkwasny-ibm, line 200: Pedantic: Not required, but some assertion to ensure the two elements in the list are the correct ones would make this test more robust.

dkwasny-ibm (approved):

PR #171, opened 2021-09-20
Add full parameter support to the CqlEvaluator

csandersdev:
I used the existing classes from from the cql-engine project. I didn't
want to link in the classes from the cql-engine directly because that
project has a number of HAPI FHIR dependencies that we don't want to
depend on in the spark application. I could certainly jump through some
hoops with dependency management, but I think this is the best plan for
now. Long-term plan I want to migrate all of the functionality from the
cql-engine to new projects like this one (probably at least one more for
measure-evaluator) and then do away with that project.


dkwasny-ibm (approved):
Looks good to me

PR #173, opened 2021-09-21
Switch DataRowRetrieveProvider to use Code equivalence checking

csandersdev:
For retrieves that are filtered by codes, the DataRowRetrieveProvider
indexes the codes in the _codePath_ column on first access. The CodeKey
object is used as the index key. CodeKey extends the CQL Code object and
implements hashCode and equals so it works as a Map key. It *was* using
the CQL Engine's Code.equal method for equality checks. Code.equal
takes into account all fields in the Code including version and display.
Per the [CQL spec](https://cql.hl7.org/02-authorsguide.html#filtering-with-terminology), the retrieve filtering should use equivalence checking
instead. I updated CodeKey.equals to use Code.equivalent and fixed the
hashCode method so that it only considers the system and code fields.

I made a small change to the DataRowRetrieveProvider to handle null
values and to allow users to specify any type of column in codePath. If
the column is non-null and not already a Code, a Code will be created
and the value will be set to the toString of the codePath object.

The DataRowRetrieveProviderTest was refactored to avoid a bunch of
unnecessary duplication of test data initialization. I then fleshed out
the tests so that we have complete code coverage of all the code paths.

I also added a unit test to SparkDataRow for the case where column
metadata is used and the user specifies a default system for all
codes in that column.

### DIFF
--- a/cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/data/SparkDataRowTest.java
+++ b/cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/data/SparkDataRowTest.java
@@ -32,6 +32,7 @@ import com.ibm.cohort.cql.spark.BaseSparkTest;
 import com.ibm.cohort.datarow.exception.UnsupportedConversionException;
 
 public class SparkDataRowTest extends BaseSparkTest {
+    private static final String SNOMED = "http://snomed.info/sct";
     private static final long serialVersionUID = 1L;
 
     @Test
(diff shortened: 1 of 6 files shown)


cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/data/SparkDataRowTest.java
```diff
@@ -241,10 +242,31 @@ public void testConversionSemanticsAsCodeAllFields() {
...
+        CodeWithMetadataPOJO pojo = new CodeWithMetadataPOJO("123", null, null);
+        
+        SparkDataRow sdr = runMetadataTest(pojo, codeMetadata);
+        
+        Object converted = sdr.getValue("code");
+        Code code = (Code) converted;
+        assertEquals( pojo.getCodeStr(), code.getCode() );
+        assertEquals( SNOMED, code.getSystem() );
+        assertEquals( null, code.getDisplay() );
+        assertEquals( null, code.getVersion() );
```
dkwasny-ibm, line 263: Pedantic: `assertNull()` is ever so slightly cleaner here.

dkwasny-ibm (approved):
Seems good.  Nothing but pedantic stuff from me.

PR #174, opened 2021-09-24
Wffhcohort 499 - Output parquet formatted results

msargentibm:
I'll preemptively put out an offer to do a code walk through with anyone who would like one.

Please don't let the number of files touched in this PR deter you. Many files  in the PR are related to testing.

### Primary Changes

- Change the output format of the  program from plaintext to parquet. Output results from the engine for each context as a parquet file with a context key column followed by each measure/define combination is in its own column.
- Remove UUID from the output path files when writing  results.
- Add in the option to have the program overwrite files when outputting results (the program will not overwrite results by default).
- Add in an option to control the delimiter used in creating result column names.

### Longer Explanation

The changes to  the `SparkCqlEvaluator` program are relatively small. I added in a step to pre-calculate a schema for each context and then use the schema to convert the Map of results into a dataframe for each context that is then written out as parquet. The resulting parquet files will have the very first column as the context key, and then one column per result run by the engine (currently named `MEASURE|DEFINE_NAME`).

The schema calculation work is driven by `SparkSchemaCreator`. Type information for define results is taken from the translated ELM for each library (getting this information involved changing the options used in `CqlToElmTranslator`). Context key column type information is derived by searching through the ModelInfos used in the libraries being evaluated for a context. I don't think the "happy path" for this code is all that complex,  but there are plenty of checks throughout the generation process to  try to catch cases where someone's input files are not lining up.

The majority of my tests are  for the schema generation code. I wanted to make sure I had as many edge cases as I could think of covered. I also added some validation to `SparkCohortEvaluatorTest` tests to validate the result files to varying degrees.

### Things to Keep in Mind Post-PR
- I hard-coded Decimal columns to be overly wide so that we don't accidentally lose precision when reporting  results. We should see if configurable precision is preferred. If so, we should add in the option to configure the precision of a result column in some way.
- This PR does not support running the same measure/define combination with different parameters for a given context. We will need to do work to support that use case.
- I keep flip-flopping with how I feel about the schema generation code. If anyone has ideas for improving or restructuring that portion of the code, I'd be glad  to make a ticket and partner with someone to change that bit.


cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkSchemaCreator.java
```diff
@@ -0,0 +1,202 @@
...
+			for (StructField field : resultsSchema.fields()) {
+				fullSchema = fullSchema.add(field);
+			}
+			resultsSchema = fullSchema;
+		}
+		
+		return resultsSchema;
+	}
+
+	private Tuple2<String, DataType> getKeyInformationForContext(String contextName, Set<Tuple2<String, String>> usingInfos) {
```
csandersdev, line 115: Should this be called `getDataTypeForContextKey`?

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkSchemaCreator.java
```diff
@@ -0,0 +1,202 @@
...
+		
+		if (keyType == null) {
+			throw new IllegalArgumentException(
+					"Could not locate type information for " + primaryDataType + "." + primaryKeyColumn
+							+ " in the provided ModelInfo files. Cannot infer key type for context: " + contextName);			
+		}
+		return new Tuple2<>(contextDefinition.getPrimaryKeyColumn(), keyType);
+	}
+	
+	private ContextDefinition getContextDefinition(String contextName) {
```
csandersdev, line 176: This is a good candidate for a helper method `getContextDefinitionByName` on the ContextDefinitions object.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -125,7 +129,132 @@ public void testAllTypesEvaluationSuccess() throws Exception {
...
+                ),
+                new StructType()
+                        .add("id", DataTypes.IntegerType, false)
+                        .add("Context2Id.define_date", DataTypes.DateType, true)
+                        .add("Context2Id.define_datetime", DataTypes.TimestampType, true)
+        );
+    }
+    
+    private void validateOutput(String filename, List<Row> expectedRows, StructType schema) {
+        spark = initializeSession(Java8API.ENABLED);
```
csandersdev, line 223: This one overwrites the instance-level spark session. Is this needed?

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -125,7 +129,132 @@ public void testAllTypesEvaluationSuccess() throws Exception {
...
+        // by hand and counting the unique values for each context's key column.
+        validateOutputCountsAndColumns(patientFile.toURI().toString(), new HashSet<>(Arrays.asList("pat_id", "MeasureAB.cohort")), 100);
+        validateOutputCountsAndColumns(aFile.toURI().toString(), new HashSet<>(Arrays.asList("id_col", "MeasureA.cohort")), 572);
+        validateOutputCountsAndColumns(bFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureB.cohort")), 575);
+        validateOutputCountsAndColumns(cFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureC.cohort")), 600);
+        validateOutputCountsAndColumns(dFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureD.cohort")), 567);
+    }
+
+    private void validateOutputCountsAndColumns(String filename, Set<String> columnNames, int numExpectedRows) {
+        SparkSession sparkSession = initializeSession(Java8API.ENABLED);
```
csandersdev, line 148: We have a Spark session that is initialized in the @Before method. Do we need to get another one here or should we just be reusing the one that we already have?

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -386,34 +422,36 @@ protected CqlEvaluationRequests readJobSpecification(String path) throws Excepti
...
+                            for (String field : Arrays.copyOfRange(schema.fieldNames(), 1, schema.fieldNames().length)) {
+                                data.add(results.get(field));
+                            }
+                            return data.toArray();
+                        })
+                        .map(RowFactory::create),
+                schema
+        );
+
+        dataFrame.write().mode(args.overwriteResults ? "overwrite" : "errorifexists").format("parquet").save(outputURI);
```
csandersdev, line 454: I think it would be good to have an `--output-format` option on the program that allows the user to control the format here.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -125,7 +129,132 @@ public void testAllTypesEvaluationSuccess() throws Exception {
...
+                new StructType()
+                        .add("id", DataTypes.IntegerType, false)
+                        .add("Context2Id.define_date", DataTypes.DateType, true)
+                        .add("Context2Id.define_datetime", DataTypes.TimestampType, true)
+        );
+    }
+    
+    private void validateOutput(String filename, List<Row> expectedRows, StructType schema) {
+        spark = initializeSession(Java8API.ENABLED);
+        Dataset<Row> actualDataFrame = spark.read().parquet(filename);
```
csandersdev, line 224: If we add an `--output-format` option to the main program, we will need to be able to control for that here as well.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -125,7 +129,132 @@ public void testAllTypesEvaluationSuccess() throws Exception {
...
+        // Column names with a dot in them need escaped with backticks
+        List<String> columnList = Arrays.asList(actualDataFrame.columns()).stream()
+                .map(x -> "`" + x + "`")
+                .collect(Collectors.toList());
```
csandersdev, line 228: Do you think we should change our default column name encoder to use a different delimiter than period? Perhaps a dash would be better? Maybe a vertical bar / pipe?

csandersdev (commented):
Lots of great work here, but I added some fairly minor suggestions.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkSchemaCreator.java
```diff
@@ -0,0 +1,202 @@
...
+	public StructType calculateSchemaForContext(String contextName) throws Exception {
+		List<CqlEvaluationRequest> filteredRequests = requests.getEvaluations().stream().filter(r -> r.getContextKey().equals(contextName)).collect(Collectors.toList());
+
+		StructType resultsSchema = new StructType();
+
+		Set<Tuple2<String, String>> usingInfo = new HashSet<>();
+
+		for (CqlEvaluationRequest filteredRequest : filteredRequests) {
+			CqlLibraryDescriptor descriptor = filteredRequest.getDescriptor();
+			String measureName = descriptor.getLibraryId();
```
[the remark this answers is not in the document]
msargentibm: Sure. I can rename it.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkSchemaCreator.java
msargentibm: Sure

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkSchemaCreator.java
msargentibm: Thanks for the suggestion

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
msargentibm: The spark session  gets closed when we hit the end of SparkCohortEvaluator. Calling  `initializeSession` here makes sure we re-create a session when needed, or snag the cached session.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
msargentibm: This might be a good item to create a ticket for once we have a use case to write out results using an alternate format.

I know something like delta lake, the most basic write is just swapping out the format string. But I think that there is potentially extra configuration a user would want to be able to provide if writing out results in delta format and I think fully  supporting that would be a nice sized task on its own.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
msargentibm: No strong feelings here, especially since I thought we were shortly going to allow users to specify output column names. Although I suppose we still might want a  default behavior if they don't specify anything.

Think a swap  is still worth it?

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -102,6 +106,24 @@
      * discussion in the libraryProvider documentation for complete reasoning.
      */
     protected static ThreadLocal<CqlTerminologyProvider> terminologyProvider = new ThreadLocal<>();
+    
+    protected Map<String, StructType> calculateSparkSchema(List<String> contextNames, ContextDefinitions contextDefinitions) throws Exception{
+        CqlLibraryProvider libProvider = SparkCqlEvaluator.libraryProvider.get();
+        if (libProvider == null) {
+            libProvider = createLibraryProvider();
```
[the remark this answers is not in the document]
msargentibm: I can look at this again. I know I  was having some trouble with creating the thread local since they are static, but some of them were relying on  state from the actual arguments used to  run the program.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -386,34 +422,36 @@ protected CqlEvaluationRequests readJobSpecification(String path) throws Excepti
...
-        
-
         if (args.outputPartitions != null) {
             resultsByContext = resultsByContext.repartition(args.outputPartitions.intValue());
         }
-        String uuid = UUID.randomUUID().toString();
-        LOG.info("Batch UUID " + uuid);
-        resultsByContext.saveAsTextFile(outputURI + uuid);
+
+        Dataset<Row> dataFrame = spark.createDataFrame(
```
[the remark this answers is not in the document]
msargentibm: Moved.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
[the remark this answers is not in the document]
csandersdev: Now that you say it, that makes sense. We use the input args to the program for both initializers. Probably not worth spending any more cycles worry about it.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
csandersdev: So, maybe a better suggestion here is to wrap your initializeSession in a try-with-close block? That was my first thought before I remembered we had the annotated methods doing initialization and teardown. It probably doesn't matter much in the end because these are unit tests, but I try to make sure to always close resources when I'm done.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
csandersdev: My main concern here is that we know our first consumer is going to use delta lake on input and it follows that they will probably expect delta lake on output. I agree that there are other things a user might want to control, but the format itself is easy to add and makes sure delta is available to the end user. As-is, they can't control the output format. I would also be here with simply removing the `.format("parquet")` specifier and allowing the user to use `spark.sql.sources.default` as the control point. That was how I initially did the read portion of the app before adding the `--input-format` option. I think it is best to add `--output-format` though.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
csandersdev: I think there is a place for a default name encoder even when we add the ability for users to specify their own column names. The default encoder will need to get smarter so that it handles duplicates properly and maybe we also add a configuration option for column name delimiter there as well? That would alleviate from us the responsibility of choosing a universally acceptable value.  I just know that if you hit a problem with the column names, then the next person to use our solution is almost certainly going to hit the same problem. There is a workaround, so it isn't urgent. I'm just trying to make our stuff as easy to use as possible.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
msargentibm: >My main concern here is that we know our first consumer is going to use delta lake on input and it follows that they will probably expect delta lake on output

I understand this thought, but I think writing out delta format from our program runs counter to the goal of having users write to a particular output location once.  If we write out delta tables, then I feel like that implies that you can re-run the program and append results to a previously created table since that is the main reason you would use delta lake as an output format.

If we want to signal that we only support writing  to separate locations or  overwriting previous results, then not having delta lake as an option probably gets this point across better.

dkwasny-ibm: I agree with the concern of DeltaLake running counter to our usecase.
No need to take on that extra dressing if we don't actually plan to use it (i.e. persistent results that get updated).
I could possibly see users wanting us to write using other "one-shot" formats like `csv` in the future, but I feel like we could add that later pretty easily.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -125,7 +128,153 @@ public void testAllTypesEvaluationSuccess() throws Exception {
...
+        validateOutputCountsAndColumns(patientFile.toURI().toString(), new HashSet<>(Arrays.asList("pat_id", "MeasureAB|cohort")), 100);
+        validateOutputCountsAndColumns(aFile.toURI().toString(), new HashSet<>(Arrays.asList("id_col", "MeasureA|cohort")), 572);
+        validateOutputCountsAndColumns(bFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureB|cohort")), 575);
+        validateOutputCountsAndColumns(cFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureC|cohort")), 600);
+        validateOutputCountsAndColumns(dFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureD|cohort")), 567);
+    }
+
+    private void validateOutputCountsAndColumns(String filename, Set<String> columnNames, int numExpectedRows) {
+        // SparkCqlEvaluator closes the SparkSession. Make sure we have one opened before any validation. 
+        try(SparkSession sparkSession = initializeSession(Java8API.ENABLED)) {
```
dkwasny-ibm, line 148: Unless there's a deeper reason, I say we just let the `SparkSession` objects live.
`SparkSession.Builder` should take care of reusing sessions for us and the more sessions we create, the slower our overall test suite is.

dkwasny-ibm (changes requested):
Looks good overall.  I have some small requests I'd like addressed or explained, but not much else.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkSchemaCreator.java
```diff
@@ -0,0 +1,191 @@
...
+										 .collect(Collectors.toList()));
+
+				List<ExpressionDef> expressionDefs = library.getStatements().getDef().stream()
+						.filter(x -> x.getName().equals(expression))
+						.collect(Collectors.toList());
+
+				if (expressionDefs.isEmpty()) {
+					throw new IllegalArgumentException("Expression " + expression + " is configured in the CQL jobs file, but not found in "
+															   + descriptor.getLibraryId() + "." + descriptor.getVersion());
+				}
```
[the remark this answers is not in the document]
msargentibm: I  had this check previously, but removed it because the cql translator wouldn't let me translate cql with duplicate define names. I can add the check back just in case the translator behavior changes at some point.

shelleyMurphy (approved):

deekim (approved):
LGTM.
Thanks for the detailed write-up.
FWIW, I think that `parquet` makes sense for now, and the "user-controlled" `output-format` can be a separate ticket.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
dkwasny-ibm: Talked with Matt outside of this PR.  I don't think any changes are necessary here.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
csandersdev: The group consensus appears to be to keep hard-coding parquet for now. I turned my feedback here into a JIRA ticket (WFFHCOHORT-617) and will resolve the conversation so that the PR can move forward.

dkwasny-ibm (approved):

csandersdev (approved):
Feedback has been addressed. LGTM

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
msargentibm: Maybe let's ask HI if they specifically want delta tables as an output from our program. That should make it pretty easy to see if this is something worth adding or not.

PR #177, opened 2021-09-29
Add additional error handling and test cases

csandersdev:
* Created a cleaner error message in SparkSchemaCreator for cases where
the getResultTypeName() for a function returns null. This is
particularly relevant when the cohort returns a list.
* Added error handling to SparkSchemaCreator for when a requested
library is not availble in the CqlLibraryProvider (it returns null).
* Added exception handling around the evaluator.evaluate call in
SparkCqlEvaluator so that the context information is captured for the
job that fails (context, context key, etc.).


msargentibm (approved):
Thanks for the enhancements.

PR #178, opened 2021-09-30
Wffhcohort 621

msargentibm:
Removing old AWS classes. Also updated some poms to git rid of declared unused dependencies.

I was basing my search on "aws" in our repo, but please let me know if I missed anything.


cohort-evaluator-spark/pom.xml
```diff
@@ -98,11 +98,6 @@
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
 		</dependency>
-
-		<dependency>
-			<groupId>com.amazonaws</groupId>
-			<artifactId>aws-java-sdk-bundle</artifactId>
-		</dependency>
```
csandersdev, line 101: We need to check the Dockerfile and make sure that the AWS JAR is still included in the build before we merge this down. Theoretically, the `hadoop-aws` dependency will pull it in, but I'm not sure about that because hadoop-aws is tagged as a runtime dependency. None of our unit tests connect to S3, so we wouldn't find out about a missing JAR until we tried to run this under Docker.

msargentibm: Just did a clean build, ran the Dockerfile for cohort-evaluator-spark, and then exec'd into a running docker container made from the image. I am seeing `aws-java-sdk-bundle-1.11.375.jar ` listed in `/opt/spark/jars`.

I'd be happy for someone else to sanity check this result as well.

dkwasny-ibm (approved):

cohort-evaluator-spark/pom.xml
dkwasny-ibm: A quick run of the job will confirm whether the right jars are there.

csandersdev: I pulled the branch, built the docker image, and ran `docker run -it --entrypoint /bin/bash tag:version` and I see the AWS JAR in /opt/spark/jars. I think we are good here.

csandersdev (approved):
LGTM

PR #179, opened 2021-10-04
Initial Spark User Guide

csandersdev:
This is an initial draft of a user guide for the spark application. It is absolutely not intended to be the final form, but I wanted to get something out that folks could review sooner rather than later.


docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+Spark configuration itself can be quite complex and there are a number of configuration options assumed to in place for the above example to function. These properties could be specified as `--conf` arguments to spark-submit or our preference is to store these in a separate properties file such as SPARK_HOME\conf\spark-defaults.conf. This guide makes no attempt to cover the complexity of Spark configuration. Necessary properties will vary based on the deployment environment and application needs. For reference, the full set of Spark properties available can be found [here](https://spark.apache.org/docs/latest/configuration.html#spark-configuration). Details on how to configure Spark to run on Kubernetes can be found [here](https://spark.apache.org/docs/latest/running-on-kubernetes.html).
```
rickjstevens, line 24: ...assumed to "be" in place...

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
+
+# CQL Evaluation using Apache Spark
+
+## Overview
+
+The CQL evaluation function provided by the quality-measure-and-cohort-service project can be deployed in a variety of modes depending on the needs of the consuming solution. The most flexible option is a client-server mode of operation where the client performs the CQL evaluation for data that is stored in any remote FHIR server supporting the FHIR REST specification. While flexible, the client-server mode is likely not ideal for applications that need to process very large datasets due to the amount of data serialization and transfers required. For solutions wanting to process big data there is program provided that runs under Apache Spark. In this mode, the FHIR data model is eschewed in favor of traditional tabular data used by many analytics applications. Facilities are provided for loading the data from randomly ordered files, aggregated based on a specific execution context (e.g. Patient, Claim, etc.), and then evaluated using any number of CQL inputs. Outputs are written back out as tabular data. This guide describes how to configure and use that application.
```
rickjstevens, line 6: Would it be good to include a variation of the diagram we have showing various ways to run our CQL evaluation engine? One showing both the FHIR use case and Apache Spark use case, but maybe not the IBM FHIR embedded option?

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Program assets
+
+The cohorting capabilities (aka [quality-measure-and-cohort-service](https://github.com/Alvearie/quality-measure-and-cohort-service/)) are delivered as a public open source project under the Alvearie organization in Github. The module of interest for users consuming the Apache Spark function is  [cohort-evaluator-spark](https://github.com/Alvearie/quality-measure-and-cohort-service/tree/cql-spark-evaluator). Provided inside that project is a sample [Dockerfile](https://github.com/Alvearie/quality-measure-and-cohort-service/blob/cql-spark-evaluator/cohort-evaluator-spark/Dockerfile) that can be built and published to the appropriate place in your execution environment. No publicly published version of the docker image exists at this time.
```
rickjstevens, line 10: Note for future reference - we'll need to update this paragraph once we have publicly available docker image of our cohort spark application available - which should be on our roadmap, pending toolchain function updates.

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
... it is important that the fields of the model that contain code data be defined as the System.Code datatype in the modelinfo file and that the CQL engine be able to fill in the extra code-related fields as needed. Support for this has been made available via metadata fields in the Spark schema. A column can be marked `isCodeCol = true` in the metadata and the column contents will automatically be converted into a System.Code value. Additionally, the metadata can contain `system = my-uri-value` which will assign a default codesystem URI to each System.Code that is created or `systemCol = other-column-name` can be used to retrieve the codesystem URI from another column in the source data. There is also a `displayCol = other-column-name` option that controls setting the display text for a code. No metadata is provided for codesystem version. Best practices established by the CQL author team recommend that users right CQL using the tilde (~) operator for code comparison which only considers code and codesystem URI values in the comparison. This is the default behavior of the [Filtering with Terminology](https://cql.hl7.org/02-authorsguide.html#filtering-with-terminology) functionality.
```
rickjstevens, line 56: ...users "write" CQL using...

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+### CQL Job Descriptions
+
+The CQL jobs that will be evaluated by a run of the Spark CQL Evaluator application are stored in a JSON-formatted configuration file that is typically called cql-jobs.json. Again, the name isn't prescribed, but the rest of this guide will use the name cql-jobs.json for simplicity. The cql-jobs.json file lists each CQL Library that will be used, which expressions (aka define statements) within the CQL Library will be evaluated, and any input parameters that will be passed to the CQL evaluation engine to be used by parameterized logic within the library, and a `contextKey` that links the CQL to a particular aggregation context defined in the `context-definitions.json` file. A convenience item is provided for configuring CQL parameters at a global level. This is potentially useful for something like a measurement period that is intended to be constant throughout the evaluation of all libraries and statements being evaluated.
```
rickjstevens, line 101: Maybe there's a good reason we did it this way - or maybe this is something we could consider as a future enhancement, but given a CQL library and a define statement, could we not determine the context that is in effect for that define statement, versus having it explicitly called out in cql-jobs?

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+If the user wishes to limit which jobs are evaluated during a specific application run, there are program options available that limit which libraries (`-l`) and expressions (`-e`) are evaluated. For the purposes of job filtering, cql library version is ignored. If only the `-e` option is provided, then all expressions with that name will be evaluated across all cql job definitions. 
+
+### Program Output
+
+For each aggregation context, an separate output table is created. The storage path for the output table is configured using `-o` options as described above. The records of the context output table consist of the unique key for the specific context record (sharing the same name as the field in the model definition) and a column for each CQL expression that is evaluated. The default column name for the CQL evaluation columns is "<Library Name>|<Expression Name>". The hyphen delimiter can be changed using the `--default-output-column-delimiter` program option or users can provide their own column name alongside the expression in the `cql-jobs.json` file. If more than one column ends up with the same name, a runtime error will be thrown.
```
rickjstevens, line 147: ...each aggregation context, "a" separate...

rickjstevens, line 147: It wasn't clear to me how you can express a custom output column name alongside an expression in cql-jobs.json. Maybe add that to the example and elaborate a bit more on how that's done?

rickjstevens (approved):
This looks really good - just a few typo and items to clarify a bit more.

docs/user-guide/spark-user-guide.md
csandersdev: I was intentionally fudging here because that work is still in progress. I debated leaving it out entirely, but I suspect it will be there by the time HI actually starts using our solution. The actual work will probably involve a change to cql-jobs.json that looks like this... `expresssions: [ { "name": 'MyDefine', outputColumn: 'OutputColumnName' } ]` but I wasn't willing to commit to that until Matt finished the work.

docs/user-guide/spark-user-guide.md
csandersdev: A CQL library can technically have 0..n `context` statements (e.g. CQF Measures IG [here](https://hl7.org/fhir/us/cqfmeasures/2021May/measure-conformance.html#patient-based-calculation-1)). Even if we assumed that each library had exactly one `context` statement, I like that the current design is flexible enough to all the user to execute the same library across different *aggregation* contexts. For example, one job might join patient and eligibility and another job might join patient and claims data. Both are "context Patient" from the perspective of the CQL, but the underlying data we collect is different. I don't know how much people would actually use this, but it seems plausible to split up jobs this way. I'd be willing to revisit if there is a strong feeling that we should limit the CQL to exactly one aggregation context.

docs/user-guide/spark-user-guide.md
csandersdev: Agree that we need a solution. I debated leaving this out entirely, but I tried to create the documentation with both internal and external users in mind, so I thought it good to at least make a statement. We can certainly revise later.

docs/user-guide/spark-user-guide.md
csandersdev: In my earliest draft of this paragraph I had assumed that we would put that diagram here, but I decided against it later because I think it focuses too much on things that aren't relevant to this guide specifically. I think we should have another page in the user guide that talks about the solution as a whole, includes that diagram, and links to this page for the specifics on Spark usage. I'd like to keep this page focused on the Spark implementation.

docs/user-guide/spark-user-guide.md
msargentibm: The only other option I thought of is a separate field with mappings. Something like `outputColumnOverride: [{"expression1": "outputName1"}, {"expression2": "outputName2"}]`. Main advantage would be that existing tests and cql jobs files should be backwards compatible with adding a completely separate field that controls output name overrides. That being said, I think I like the logical grouping that you get from using the example format that Corey shared.

Does anyone have strong preferences one way or another?

csandersdev: I was thinking about backwards compatibility myself and I wonder if we could implement a custom deserializer for that field that understands how to deal with both Strings and Maps? Seems plausible.

In the same vein, I don't really like putting outputColumnName in the CqlEvaluationRequest object because outputColumnName is specific to the Spark application and my original design goal was to keep that object generic enough to be used elsewhere, such as if/when we update the CLI to use the cql-evaluator project. If the fields were separate, you could view that as an extension to the base object  (e.g. SparkCqlEvaluationRequest) .

Would we ever put the column name overrides as a single top-level element in CqlEvaluationRequests instead of CqlEvaluationRequest? Or does it belong in a completely separate overrides file? I'm not sure either is a good idea from a usability standpoint, but I like the idea of having all of the overrides for a single context in a single place because it helps make it clear that they are all related and that column names need to be unique.

msargentibm: > Would we ever put the column name overrides as a single top-level element in CqlEvaluationRequests instead of CqlEvaluationRequest?

I'd have to think about this a bit. I think  if we separated out the overrides completely, then we'd need a unique name/id per CqlEvaluationRequest so we could tie overrides to a particular request.

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+The entry point to the Spark application is the `com.ibm.cohort.cql.spark.SparkCqlEvaluator` class. The typical mode of invocation is to use the appropriate spark-submit script from SPARK_HOME\bin to execute the application. The simplest invocation for a Windows user will look like this...
+
+```
+%SPARK_HOME%\bin\spark-submit.cmd --name cohort-evaluator-spark --class com.ibm.cohort.cql.spark.SparkCqlEvaluator local:/opt/spark/jars/cohort-evaluator-spark/target/cohort-evaluator-spark-1.0.2-SNAPSHOT.jar @/path/to/program-args.txt
+```
+
+In this example, the `@/path/to/program-args.txt` is a convenience option that allows users to store their program arguments in a file that can be reused across multiple program executions. Each line of the target file is a separate positional parameter. The option name and option value (for those options that have a value) should be on separate lines. Make sure to avoid trailing whitespace as that can confuse the argument parser. Alternatively, you can just pass all the options (e.g. -c /path/to/cql -m /path/to/modelinfo.xml -d /path/to/context-definitions.json -j /path/to/cql-jobs.json ...) directly in the spark-submit execution.
```
jaigoradia, line 22: May be need to provide example of output from usage statement to show the list of program-args that are available.

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
...om a number of files that work together to describe the desired program behavior. The content for each of those configuration files will be discussed in detail below, but, in summary, the configuration options are the path to a filesystem folder containing files in either CQL or ELM+XML format (`-c`), the path to a CQL modelinfo file in XML format (`-m`), and paths to a context-definitions.json file that describes how Spark data will be grouped prior to CQL evaluation (`-d`) and a cql-jobs file that describes which CQL libraries and expressions will be executed during evaluation (`-j`). All *configuration data* is assumed to be available in the local filesystem where the Spark application is running. For a Spark cluster running under Kubernetes, this can be done using a Persistent Volume (PV)/Persistent Volume Claim (PVC). In the IBM cloud, we test with a PVC that is backed by a PV that resides in IBM Cloud Object Storage (COS). The IBM Cloud team has an excellent [tutorial](https://www.ibm.com/cloud/blog/kubernetes-persistent-volumes-backed-by-ibm-cloud-object-storage-buckets) that describes how to setup your COS-backed PVC that you should review if you wish to pursue this option.
```
jaigoradia, line 42: Somewhere here it would be helpful to insert some kind of overall architectural diagram.

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+The data being read can be in any format that Spark understands, but all data is assumed to be provided in a single, consistent format. The default format for Spark is parquet. Users can adjust the input format by setting the `spark.sql.datasources.default` configuration property when invoking the Spark application or by using the `--input-format` option to the application. The sample Dockerfile that is provided has support for all the basic formats supported natively by Spark and also Deltalake format. Users that want additional formats, such as Apache Iceberg, can use the `--packages` option to spark submit or extend the Docker image to add the additional needed JARs.
+
+Output is configured almost identically to input except it uses `-o` options and the key in the key=value pairs is the context name of a context aggregation defined in the `context-definitions.json` file. More on that later. Output format is, again, available in any supported spark format as specified by the `--output-format` program option, whatever is configured in the `spark.sql.sources.default` Spark configuration, parquet if no other configuration is specified.
```
jaigoradia, line 52: Output format is, again, available in any supported spark format as specified by the `--output-format` program option, whatever is configured in the `spark.sql.sources.default` Spark configuration, parquet if no other configuration is specified.

Should this read as: "Output format is, again, available in any supported spark format as specified by the `--output-format` program option or whatever is configured in the `spark.sql.sources.default` Spark Configuration. The output format is parquet if no other configuration is specified. ?

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Context Definitions
+
+The data tables that are used as input to the Spark CQL Evaluator application are are assumed to be large, unsorted datasets. The first step to CQL evaluation is loading the datasets and then grouping records together into related data. The CQL specification calls these groupings contexts (e.g. context Patient). Any number of evaluation contexts are supported, but each context causes data to be reread from storage and regrouped. These are potentially very expensive operations, so care should be taken when deciding on the aggregation contexts that will be used.
```
jaigoradia, line 60: two 'are's in the first statement.

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+					"relatedDataType": "Patient",
+					"relatedKeyColumn": "patient_id"
+				}]
+        }]
+}
+```
+
+For the initial pilot implementation of the Spark application, join logic is limited to the one-to-many and many-to-many scenarios described above. More complex join logic involving compound primary keys (more than one column required to uniquely identify a record) or related data beyond a single table relationship (with or without an association table in between) are unsupported. These will come in a future release.
+
+If the user wishes to limit which aggregation contexts are evaluated during an application run, there is a program option (`-a`) that can be specified one more more times that lists the contextKey value for each aggregation context that should be evaluated.
```
jaigoradia, line 97: there is a program option (`a`) that can be specified one or more times ...

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+If the user wishes to limit which jobs are evaluated during a specific application run, there are program options available that limit which libraries (`-l`) and expressions (`-e`) are evaluated. For the purposes of job filtering, cql library version is ignored. If only the `-e` option is provided, then all expressions with that name will be evaluated across all cql job definitions. 
+
+### Program Output
+
+For each aggregation context, a separate output table is created. The storage path for the output table is configured using `-o` options as described above. The records of the context output table consist of the unique key for the specific context record (sharing the same name as the field in the model definition) and a column for each CQL expression that is evaluated. The default column name for the CQL evaluation columns is "<Library Name>|<Expression Name>". The hyphen delimiter can be changed using the `--default-output-column-delimiter` program option or users can provide their own column name alongside the expression in the `cql-jobs.json` file. If more than one column ends up with the same name, a runtime error will be thrown.
```
jaigoradia, line 147: Isn't that the 'pipe' delimiter and not hyphen as in '-'?

docs/user-guide/spark-user-guide.md
```diff
@@ -0,0 +1,159 @@
...
+
+Depending on a number of factors including the amount of data in the input, how the input data is organized, and how spark parallelism is configured, output tables might end up with a large amount of partitions yielding a large number of output files. In these cases, it might be desirable to compact data prior to writing it to storage. A program option `-n` is provided that, when specified, will cause output data to be compacted into the specified number of partitions.
+
+### Debugging
+
+Sometimes errors arise during CQL evaluation that can be challenging to debug simply by reading the provided exception message. When this occurs, it is often useful to enable the CQL Engine trace facility using the `--debug` program option. This is most helpful with the `enable-locators` option is specified during CQL translation, so take care to provide that option if you are compiling the CQL yourself. CQL trace messages are written to stdout with no additional configuration options available for controlling output location available at this time.
```
jaigoradia, line 159: This is most helpful with when the `enable-locators` option is specified ...

jaigoradia (commented):
Submitted comments to correct typos/grammar at certain places and suggestions to include list of available program args and possibly adding a simple architectural diagram (if possible) showing different pieces of the puzzle in the flow.

jaigoradia (approved):
Approved.

PR #180, opened 2021-10-05
Added new fvt test files and updated run-fvttests.sh script in support of running Spark based fvt tests

jaigoradia:
General description:
This PR is to integrate support for executing Spark-submit and run Spark based tests as part of  FVT test automation in the CD toolchain. The logic that was present in run-fvttests.sh script to run REST API and CLI tests was enhanced to also support the ability to run Spark-submit command that takes in configuration information for cql evaluation from a COS bucket 'Cohort-config' and data from a folder called fvt-input-data under the encrypted COS bucket 'cohort-data-tenant2',  evaluates the cql against the data set (parquet files) based on the config information and writes out the output as parquet files to a folder fvt-output under the same COS bucket. 

The high level flow can be described as follows: (**Note the newly added section for Spark based tests is located in the lower third of the run-fvttests.sh script and book ended by an eye-catcher comment to isolate it from previous non spark based tests)

1. run-fvttests.sh script prepares a pod deployment yaml file and spins of a new pod "engine-spark-fvt-test" that uses an image that has already been pushed to the ibm container registry. This image was built from a docker file which is very similar to the one under cohort-evaluator-spark project in github and has all the spark trappings necessary. The pod has a volume mount for the 'cohort-data-tenant2' bucket in COS.
2.  Arguments for the spark-submit command to be run are built as a string object.
3.  Once the pod is running, kubectl command is used to exec into the running pod and the spark-submit command is run supplying the arguments string.  
4.  Spark does its magic from then on and spawns off the driver and executor pods. does the work to evaluate the cql and write out the output back to COS.
5. Check is made for the status of the spark-submit job to verify it ended successfully.
6. Next step is to validate the results and report number of passing/failing tests to DevOps Insights Console. The following actions are done in sequence in the run-fvttests.sh script for this purpose. 
   a. The PySpark package is installed as it is used by the validation script for comparing actual and expected results in dataframes.
   b. kubectl exec command is used to copy the entire fvt-output folder locally from the COS volume mount in the pod so that the validation script, called next, has direct access to the output parquet files without having to make separate s3a calls into COS to read them.
   c. run-fvttests.sh script calls a python script called validateSparkFvtOutput.py to do the results validation and to write out the status of passing and failing tests in the form of a junit style xml file.
   d. The generated sparkfvttest.xml file is copied into a Results folder where all other fvt tests that have been run prior also have their own xml files copied.
   e. Finally an xmlCombiner script combines output from all the xml files present in the Results folder into a single fvttest.xml file which is uploaded by the toolchain code (representing fvt test status) to DevOps Insights console. 

**The validation is done for the output from evaluating cqls  against the provided data for 4 separate contexts: Device, Observation, Patient and Practitioner. The Patient context involves one to many and many to many join logic.   Representative FVT test data and Cql logic was provided by Kwas.  Note: Validation of results for each context is considered 1 FVT test (so 4 tests overall) as far as reporting to DevOps Insights Console.


tests/check-spark-submit-status.sh
```diff
@@ -0,0 +1,42 @@
...
+
+#!/bin/bash
+
+SPARK_POD_NAME=$1
+DVRPODPREFIX="cohortfvt-spark"
+driverpod=`kubectl -n ${CLUSTER_NAMESPACE} get pods --no-headers -o custom-columns=":metadata.name" | grep ${DVRPODPREFIX} | grep driver`
+echo "Driver pod is: ${driverpod}"
+echo "Getting status of driver pod: ${driverpod}"
+status=`kubectl -n ${CLUSTER_NAMESPACE} get pod ${driverpod} --no-headers -o custom-columns=":status.phase"`
+echo "driver pod status is: ${status}"
```
csandersdev, line 15: Capitalize "driver" so that it is consistent with the above log messages?

tests/check-spark-submit-status.sh
```diff
@@ -0,0 +1,42 @@
...
+count=1
+while ( [ "${status}" == "Pending" ] || [ "${status}" == "Running" ] ) && [ ${count} le 30 ] 
+do
+   echo "Polling again for the status of driver pod after a wait of 10 seconds"
+   sleep 10
+   status=`kubectl -n ${CLUSTER_NAMESPACE} get ${driverpod} --no-headers -o custom-columns=":status.phase"`
+   echo "driver pod status is: ${status} - count is : ${count}"
```
csandersdev, line 22: Same capitalization comment as above

tests/run-fvttests.sh
```diff
@@ -95,7 +95,34 @@ kubectl -n ${CLUSTER_NAMESPACE} cp "${POD_NAME}:/bzt-configs/tests/results" "${O
...
+kubectl -n ${CLUSTER_NAMESPACE} delete pod ${olddriverpod}
+
+# Build up the spark-submit command arguments as a string
+sparkargs="--deploy-mode cluster --name cohortfvt-spark --class com.ibm.cohort.cql.spark.SparkCqlEvaluator  --conf spark.sql.shuffle.partitions=10 --conf spark.default.parallelism=10 local:///opt/spark/jars/cohort-evaluator-spark-1.0.2-SNAPSHOT.jar -d /cohort-config/fvt/fvt-context-definitions.json -j /cohort-config/fvt/fvt-cql-jobs.json -m /cohort-config/fvt/fvt-model-info.xml -c /cohort-config/fvt/cql --input-format delta -i Device=s3a://cohort-data-tenant2/fvt-input-data/Device -i Observation=s3a://cohort-data-tenant2/fvt-input-data/Observation -i Patient=s3a://cohort-data-tenant2/fvt-input-data/Patient -i PatientDeviceJoin=s3a://cohort-data-tenant2/fvt-input-data/PatientDeviceJoin -i Practitioner=s3a://cohort-data-tenant2/fvt-input-data/Practitioner --overwrite-output-for-contexts -o Patient=s3a://cohort-data-tenant2/fvt-output/Patient_cohort -o Observation=s3a://cohort-data-tenant2/fvt-output/Observation_cohort -o Practitioner=s3a://cohort-data-tenant2/fvt-output/Practitioner_cohort -o Device=s3a://cohort-data-tenant2/fvt-output/Device_cohort"
```
csandersdev, line 112: `spark.sql.shuffle.partitions` and `spark.default.parallelism` probably belong in spark-defaults.conf unless you plan on changing them between test runs. It isn't required, but you might consider storing your program arguments in a file that you copy into the running POD. It would greatly simplify this string.

tests/run-spark-fvt-deploy.yaml
```diff
@@ -0,0 +1,34 @@
...
+    app: POD_NAME_PLACEHOLDER
+spec:
+  restartPolicy: Never
+  imagePullSecrets:
+    - name: all-icr-io
+  containers:
+  - name: spark-fvt-test-app
+    image: us.icr.io/vpc-dev-cohort-rns/cohort-spark-fvt:1.0.0
+    command: ["sleep"]
+    args: ["28800"]
```
csandersdev, line 22: I'm confused by this sleep. What is it trying to accomplish?

tests/scripts/validateSparkFvtOutput.py
```diff
@@ -0,0 +1,231 @@
...
+
+	# write out header portion of xml file
+	xmlfile.write("<?xml version='1.0' encoding='UTF-8'?>" +"\n")
+	xmlfile.write("<testsuites>" + "\n")
+	xmlfile.write("\t"+ "<testsuite name=\"Spark Fvt Tests\">" + "\n")
+	# call individual validation functions in order
+	validateDeviceCohortEval()
+	validateObservationCohortEval()
+	validatePatientCohortEval()
+	validatePractitionerCohortEval()
```
csandersdev, line 217: There appears to be a lot of duplicated code in these four methods. If you passed in the filename and expectedData, would this be able to be a single function instead of four separate methods?

tests/src/main/resources/sparkconf/pod-cohort-evaluator-spark.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: Pod
+apiVersion: v1
+metadata:
+    name: cohort-evaluator-spark
+spec:
+    # The cohort-evaluator-spark Docker image runs the application
+    # under the "spark" user which is created with uid 185
+    securityContext:
+        runAsUser: 185
+        fsGroup: 185
```
csandersdev, line 16: This file is a duplicate of cohort-evaluator-spark/pod-cohort-evaluator-spark.yaml. It would be nice if we didn't have multiples floating around, especially since these uid/gid values aren't directly sync'ed with the Dockerfile where they are created. I'm not clear on what the best way to do that would be though. It is generally frowned upon to copy files from a separate module.

tests/src/main/resources/sparkconf/spark-defaults.conf
```diff
@@ -0,0 +1,50 @@
...
+#
+
+# Default system properties included when running spark-submit.
+# This is useful for setting default environmental settings.
+
+spark.default.parallelism           200
+spark.driver.memoryOverhead         1G
+spark.driver.memory                 1G
+spark.executor.memory               8G
+spark.master                        k8s://https://c105.us-east.containers.cloud.ibm.com:32327
```
csandersdev, line 25: I think you should have `spark.executor.memoryOverhead` set here as well. Also, I believe Kwas was able to reduce executor memory to 4G in his testing after memoryOverhead was increased. It isn't blocking at 8G, but it does use resources unnecessarily for a bit.

csandersdev (changes requested):
The only think blocking for me is the math issue in `check-spark-submit-status.sh`. Otherwise, I provided some general feedback and suggestions.

tests/run-spark-fvt-deploy.yaml
jaigoradia: That was something done so as to to be able to exec into the pod even after the run is done.  Once a pod goes into Completed status you do not have an ability to exec into the pod. This provides a window in time to be able to do that. Not sure i would need to do it but kept it there as that is what i do also for the other pod is spin of for the taurus based tests.

csandersdev (approved):
LGTM

PR #183, opened 2021-10-08
Wffhcohort 620 -- Configure output colum names

msargentibm:
NOTE: I am still working on a possible refactor of the code to leverage IDs when looking up output column names rather than entire CqlEvaluationRequest objects. I figured it was better to get the existing code in PR form so folks can at least take a look at the overall logic.

This PR introduces the option to specify an `outputName` for expressions that will be evaluated in the CQL jobs file. Expressions may now have any of the following formats:

1.  `"cohort"`
2.  `{"name":"cohort"}`
3.  `{"name":"cohort", "outputName":"colA"}`

Options 1 and 2 are equivalent and will result in output column names using the default naming scheme. Option 3 would evaluate the same define, but the result will be stored on a column named `colA` instead of the default name.

Primary changes:

- Add `CqlExpressionConfiguration` and `CqlExpressionConfigurationDeserializer` to enable the new CQL jobs syntax described above.
- Move filtering logic and parameter override behaviors earlier in the program (when files are read).
- Add classes responsible for parsing and validating output column names (mostly checking for duplicate names within each context).

As I try to refactor, it's likely that `ConfigurableOutputColumnNameEncoder` will change a bit.


cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -402,94 +523,6 @@ private CqlEvaluationRequest makeEvaluationRequest(String contextName, String li
         return cqlEvaluationRequest;
     }
 
-    @Test
```
msargentibm, line 405: Forgot to add these tests back. Will replace them soon.

msargentibm:
I tried to see if I could get Jackson so assign ids to objects when reading them, but I could only figure out how to generate ids when writing objects.

I'm not convinced the changes in 2690d8c are worth it, but would like to hear what people think.

msargentibm:
Looks like I have some tests to update.

cohort-evaluator/src/main/java/com/ibm/cohort/cql/evaluation/CqlEvaluationRequest.java
```diff
@@ -85,32 +109,4 @@ public String toString() {
...
-	@Override
-	public int hashCode() {
-		return new HashCodeBuilder(17, 37)
-				.append(descriptor)
-				.append(expressions)
-				.append(parameters)
-				.append(contextKey)
-				.append(contextValue)
-				.toHashCode();
-	}
```
dkwasny-ibm, line 115: Pedantic: I don't see us actually using `CqlEvaluationRequest` within a `Set` or as a `Map` key, but is there any reason to not keep `hashCode()` and `equals()` here?

PR #184, opened 2021-10-11
implement support for any column (external functions)

deekim:
There were three patterns from the spike that I got working to support the _"any column"_.
1. external functions pattern (`any-column-external-function` branch, _i.e._ this branch)
2. spark metadata pattern ([any-column-spark-datarow](https://github.com/Alvearie/quality-measure-and-cohort-service/pull/185) branch)
3. aggregate any columns on `HI` side (not creating a PR for this one right now because `HI` didn't really seem interested in doing this and the code change to support on our end is pretty minimal). 

The external functions patterns work without any changes to the `HI` data,
while the spark metadata pattern takes a similar approach to how the existing code converts `System.Code`s.
Because the solutions are unique, we can also support both patterns,
but I'll leave it to the group to see which pattern is preferred.

---
#### `HI` data
→  no changes

#### model-info
→  no changes

#### cql
```sql

codesystem "SNOMED": 'http://snomed.info/sct'
codesystem "OTHER": 'http://system.other'
code "MadeUpCode": 'cd1' from "SNOMED" display 'string1'
code "OtherCode": 'cdA' from "OTHER"

concept "MyConcept": {
	"MadeUpCode", "OtherCode"
} display 'MyConcept'


define function AnyColumn(key Choice<Any>, key String) returns List<Any> : external


define "Passes all checks":
  [A] typeA where
    exists(AnyColumn(typeA, 'code_col') intersect "MyConcept".codes)
```

---

#### `HI` data
* add spark column metadata to:
	* identify "primary" column to query against
	* "prefix" (or other metadata pattern) to identify additional columns

#### model-info
```xml
<element name="code_col" elementType="list&lt;System.Code&gt;" />
<element name="code_col2" elementType="System.Code"/>
```
→ `elementType` of `System.Code` becomes a `list<System.Code>`

#### cql
```sql
define "Passes all checks":
  [A] typeA where
    (exists(typeA.code_col intersect {"code1", "code2"}))
```

---

## references
- [WFFHCOHORT-449](https://jira.wh-sdlc.watson-health.ibm.com/browse/WFFHCOHORT-449)

## todo
- [x] add examples to [watson-health-cohorting/hi-pilot-measures](https://github.ibm.com/watson-health-cohorting/hi-pilot-measures)


cohort-evaluator-spark/src/test/resources/any-column/cql/MeasureAnyColumn-1.0.0.cql
```diff
@@ -0,0 +1,22 @@
...
+concept "MyConcept": {
+	"MadeUpCode", "OtherCode"
+} display 'MyConcept'
+
+context A
+
+define "cohort":
+  exists("Passes all checks")
+
+define function AnyColumn(key Choice<Any>, key String) returns List<Any> : external
```
msargentibm, line 18: Maybe a conversation for the future, but would it be useful to publish a cql file that has our current collection of external functions?  It might be useful to have all of our external functions available in one place that can be imported easily, even though we only have 2 of them at the moment.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -143,6 +143,30 @@ public void testAllTypesEvaluationSuccess() throws Exception {
...
+            "--input-format", "parquet",
+            "-i", "A=" + new File(inputDir, "testdata/test-A.parquet").toURI(),
+            "-o", "A=" + aFile.toURI(),
+            "-n", "10",
+            "--overwrite-output-for-contexts"
+        };
+
+        SparkCqlEvaluator.main(args);
+
+        validateOutputCountsAndColumns(aFile.toURI().toString(), new HashSet<>(Arrays.asList("id_col", "MeasureAnyColumn|cohort")), 572);
```
msargentibm, line 167: I like that there is this test to make sure that the external function provider is hooked in properly and that the evaluations succeed for the AnyColumn function. If you wanted bonus testing points, you could mimic `testValidateOutput()` and use a smaller set of handcrafted data to run and then verify the exact results.

You already have unit tests in `AnyColumnTest`, so it's not the end of the world to just verify the execution succeeded.

msargentibm (approved):
Nothing blocking.

If we had to pick one method, I have a slight preference for this PR. I have no qualms about merging both options down.

cohort-evaluator-spark/src/test/resources/any-column/cql/MeasureAnyColumn-1.0.0.cql
dkwasny-ibm: Yea.  Something like `FHIRHelpers`.  Maybe `CohortHelpers`.

dkwasny-ibm:
I also don't really see the need to merge down _both_ approaches.  One should be enough for now.

cohort-evaluator-spark/src/test/resources/any-column/cql/MeasureAnyColumn-1.0.0.cql
rickjstevens: I really like the idea of a CohortHelpers CQL library. I know there was some talk about creating a similar reusable CQL function library for use across Watson Health, not that those functions would necessarily go into CohortHelpers, but good to start capturing reusable functions (and logic?) in CQL libraries designed for reuse.

rickjstevens (approved):
I share Matt and Kwas' sentiment - that preference would be to go with just the external function implementation given flexibility it provides without requiring changes to underlying metadata. I think it would be fine to just merge the external function implementation for now for HI team's consumption.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/functions/AnyColumn.java
```diff
@@ -0,0 +1,27 @@
...
+
+        return dataRow.getFieldNames().stream().filter(fieldName -> fieldName.startsWith(fieldPrefix))
+            .map(dataRow::getValue)
+            .collect(Collectors.toList());
+    }
+
+    public static Object AnyColumnRegex(Object object, String regex) {
+        DataRow dataRow = (DataRow) object;
+
+        return dataRow.getFieldNames().stream().filter(fieldName -> fieldName.matches(regex))
```
csandersdev, line 23: `String.matches` will recompile the regex every time it is called, so it makes me a little nervous using it in a place that will get called millions of times. However, given that the context is matching field names and I expect those patterns to be relatively trivial, I expect the difference is fairly negligible. We might consider implementing a regex cache here if it becomes a problem.

cohort-evaluator/src/main/java/com/ibm/cohort/cql/evaluation/CqlContextFactory.java
```diff
@@ -125,14 +133,14 @@ public Context createContext(CqlLibraryProvider libraryProvider, CqlLibraryDescr
             CqlTerminologyProvider terminologyProvider, CqlDataProvider dataProvider, ZonedDateTime evaluationDateTime,
             Pair<String, String> contextData, Map<String, Parameter> parameters, CqlDebug debug)
             throws CqlLibraryDeserializationException {
-        
+
         ContextCacheKey key = new ContextCacheKey( libraryProvider, topLevelLibrary, terminologyProvider, evaluationDateTime, parameters );
```
csandersdev, line 137: Did you consider whether or not the external function provider needs to be part of the context cache key? The current implementation is stateless, so it certainly isn't going to matter for the Spark application, but the CqlContextFactory lives in cql-evaluator which is intended to be the new general use API down the line. It feels to me like it is most correct to include the external function provider in the cache key.

csandersdev (commented):
As I think about terminology support and the "any" code columns, this approach doesn't support use of any columns in filtered retrieves (e.g. `[Condition: MyValueSet]` or `[Condition: any_code_col in MyValueSet]`).  The other approach (using column metadata) would work better for that.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
deekim: addressed in 89ad0cf

cohort-evaluator-spark/src/test/resources/any-column/cql/MeasureAnyColumn-1.0.0.cql
deekim: moved to `CohortHelpers` in 3881ddb

PR #185, opened 2021-10-11
implement support for any column logic (spark metadata)

deekim:
the description is primarily in https://github.com/Alvearie/quality-measure-and-cohort-service/pull/184

Feel free to leave any comments and/or which implementation you would prefer to see.

### DIFF
--- a/cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkDataRow.java
+++ b/cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkDataRow.java
@@ -6,7 +6,9 @@
 
 package com.ibm.cohort.cql.spark.data;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


deekim:
@rickjstevens,
> are you thinking you'd merge both PRs so we deliver two options for offerings to pick from for AnyColumn support? 

I think that is one possibly. If `HI` wants something that works "now", the external function implementation would require any changes to the existing delta lake data. And then later on, if they choose to add the metadata, they can update their `cql` to use that pattern. 

> In either case, do you think there are any performance benefits of one approach over the other?

They both had comparable performance when I tested locally.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkDataRow.java
```diff
@@ -100,11 +104,29 @@ public Object getValue(String fieldName) {
             } else {
                 result = doDefaultConversion(sparkVal);
             }
+
+            if (metadata.contains(IS_PRIMARY_ANY_COL)) {
```
msargentibm, line 108: The only potential downside I see so far with metadata is that a "primary any column" can no longer be retrieved on its own. If an offering goes down the metadata approach for any columns, then we have to make sure this shift in retrieve capabilities is spelled out in our documentation.

msargentibm (approved):
My comment isn't a blocker.

dkwasny-ibm (approved):
My overall thoughts on the feature can be found on the [other PR](https://github.com/Alvearie/quality-measure-and-cohort-service/pull/184#issuecomment-942675443).

PR #188, opened 2021-10-14
WFFHCOHORT-627: Add intelligent column filtering to spark reads

csandersdev:
* Implemented a dataset filtering mechanism that uses information
collected using the DBCG ElmRequirementsVisitor to determine which
columns are used in the libraries under evaluation and then
selects only those columns during the read dataset phase of the program.
I couldn't make the expression-level filtering of dataset requirements
function correctly using this code, so requirements filtering is done at
the full library level. This could end up causing more data reads than are necessary, but is better than reading everything.
* Added a fail-safe program option for removing column filtering if necessary
* Added short-circuit logic to prevent reading in related datasets that are defined in the context definition, but not used by any of the CQL libraries.
* In order to use the DBCG code, it was necessary to update our CQL
translator dependency to version 1.5.4. This caused some ripple down
effects into the cql-translation and cql-engine projects, as well as,
the SparkSchemaCreator.
* Added missing copyright headers to a bunch of files.
* Removed the unused and poorly implemented externalId field from the
CqlLibraryDescriptor object.


csandersdev:
I've got a big update coming to this PR to add support for the Any column work. Folks should probably hold off until that arrives. Or, I can close the PR for now and reopen when it is ready again.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/optimizer/DataTypeRequirementsProcessorTest.java
```diff
@@ -0,0 +1,216 @@
...
+                
+                newPaths.forEach( (key,value) -> {
+                    pathsByDataType.merge(key, value, (prev,current) -> { prev.addAll(current); return prev; } );
+                });
+            } else {
+                System.out.println("Skipping " + cld.toString());
+            }
+        }   
+        
+        System.out.println(pathsByDataType);
```
dkwasny-ibm, line 175: Pedantic: Were these println's leftover from debugging?
I'm not the biggest fan of our tests writing to stdout, but I'm OK with it if you think they should be there.

cohort-evaluator/src/main/java/com/ibm/cohort/cql/translation/CqlToElmTranslator.java
```diff
@@ -93,6 +88,34 @@ public CqlTranslationResult translate(CqlLibrary primaryLibrary, CqlLibrarySourc
 
         return new CqlTranslationResult(translatedLibrary, dependencies);
     }
+
+    public CqlTranslatorOptions getOptions() {
+        return CqlTranslatorOptions.defaultOptions()
+                 .withOptions(CqlTranslator.Options.EnableResultTypes)
+                 .withAnalyzeDataRequirements(true)
+                 //.withCollapseDataRequirements(true)
```
dkwasny-ibm, line 96: Pedantic: Might be worth adding a comment detailing why this bit is commented out.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/aggregation/ContextRetriever.java
```diff
@@ -111,38 +111,40 @@ public ContextRetriever(Map<String, String> inputPaths, DatasetRetriever dataset
...
-                Column primaryJoinCriteria = primaryDataset.col(primaryJoinColumn)
-                        .equalTo(assocDataset.col(assocPrimaryColumnName));
-                joinedDataset = primaryDataset.join(assocDataset, primaryJoinCriteria);
-
-                Column relatedJoinCriteria = joinedDataset.col(assocRelatedColumnName)
-                        .equalTo(relatedDataset.col(relatedColumnName));
-                joinedDataset = joinedDataset.join(relatedDataset, relatedJoinCriteria);
-            } else {
-                throw new IllegalArgumentException("Unexpected Join Type: " + join.getClass().getName());
+            if( relatedDataset != null ) {
```
dkwasny-ibm, line 114: I think it's worth logging a warning that a particular datatype join didn't come back with a Dataset.
I _think_ this only happens when a user specifies a join, but never actually references the type in their CQLs.
If an unexpected null ever makes it through, the log could help indicate what happened.

csandersdev: The null is new with my change. It was added to support exclusion of dataset reads when the data in that dataset is never actually used in the CQL code. I wouldn't log it as a warning, but I have no objection to logging an INFO message here. I agree it could be useful in debugging.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/optimizer/DataTypeRequirementsProcessorTest.java
dkwasny-ibm: I'm fine leaving it if you get value out of it.  Just wanted to be sure it wasn't leftover from the initial development of the tests.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -155,6 +159,35 @@ public void testAllTypesEvaluationSuccess() throws Exception {
...
+          "-a", "Patient",
+          "-n", "1",
+          "--output-format", "parquet",
+          "--overwrite-output-for-contexts"
+        };
+
+        SparkCqlEvaluator.main(args);
+
+        validateOutputCountsAndColumns(patientFile.toURI().toString(), new HashSet<>(Arrays.asList("pat_id", "Parent|cohort")), 100, "parquet");
+    }
```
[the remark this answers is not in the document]
msargentibm: > I couldn't figure out how the client would be able to figure out that filtering happened

I was noodling on this too for a few and was also not coming up with a test that filtering happened. With this test we at least know that _if_ filtering happened that the correct columns were still retrieved.

msargentibm (dismissed):
As best as I can tell, the new code looks solid. It took a few passes over it to parse everything and realize that some larger methods weren't as gnarly as they initially seemed.

The only somewhat actionable thought I had was w.r.t. whether filtering behavior can be verified by in the SparkCqlEvaluatorTest tests.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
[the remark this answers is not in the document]
dkwasny-ibm: I think a full on test that verifies filtering correctly happened, while nice, isn't completely necessary.
Even a test that just verifies using the option results in the engine working correctly would be good.

msargentibm:
I tried to create some test data with `NullType` and `CalendarIntervalType` columns, but it doesn't look like I am able to with write out these column types with spark.
For `CalendarIntervalType` I get the  error:
> Cannot save interval data type into external storage.

For `NullType` I get:
> Parquet data source does not support null data type.

I'm not sure if there are other types to try that would be unsupported in our code.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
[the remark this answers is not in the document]
csandersdev: I added a few tests that exercise the filtering flag. They still don't prove that filtering was actually done, but they do prove that the flag affects some behavior.

cohort-evaluator/src/main/java/com/ibm/cohort/cql/translation/CqlToElmTranslator.java
csandersdev: I've left these in place with both uncommented. I believe they are no longer relevant because we are no longer using the elm-fhir project and that is the only thing that references them, but I need to do some investigation to confirm that. I'm going to poke Bryn and see if he can clarify.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/optimizer/PathCaptureVisitor.java
```diff
@@ -0,0 +1,125 @@
...
+        try {
+            return super.visitLetClause(elm, context);
+        } finally {
+            context.getCurrentQueryContext().exitLetDefinitionContext();
+        }
+    }
+
+    @Override
+    public Object visitExpressionDef(ExpressionDef elm, C context) {
+        LOG.trace("--> {}:{}", getCurrentLibraryIdentifier().getId(), elm.getName());
```
dkwasny-ibm, line 83: Pedantic: Would it make sense to have these "in n out" log statements for the other scope types, or do we only really care about `ExpressionDef` for now?

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/optimizer/QueryContext.java
```diff
@@ -0,0 +1,80 @@
...
+    public QueryAliasContext resolveAlias(String aliasName) {
+        if( definitionContext != null && aliasName.equals(definitionContext.getAlias()) ) {
+            return definitionContext;
+        } else {
+            return aliases.get(aliasName);
+        }
+    }
+
+    public QueryLetContext resolveLet(String letName) {
+        return letContexts.get(letName);
```
dkwasny-ibm, line 78: Will we ever attempt to resolve a `let` before the call to `exitLetDefinitionContext()` and need a check similar to `resolveAlias()`?
Does the nature of CQL require exiting a `let` before we can actually reference it?
My peabrain is struggling to follow all of this to find the answer myself.

dkwasny-ibm (changes requested):
Nothing major at all.  Just some small suggestions and questions.
The pedantic stuff can be ignored if desired.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/optimizer/ExpressionContext.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+    
+    public VersionedIdentifier getLibraryIdentifier() {
+        return libraryIdentifier;
+    }
+    public ExpressionDef getExpression() {
+        return expression;
+    }
+    
+    private Stack<QueryContext> queryContextStack = new Stack<QueryContext>();
```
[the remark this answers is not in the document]
csandersdev: This was a result of copy/paste job from the OSS code. I typically try to avoid any synchronized collection when possible, but I didn't notice that this one creeped in. There is a [slight behavior difference](https://www.baeldung.com/java-deque-vs-stack) between Stack and Deque related to iteration, but I think Deque does it the way that you'd actually want anyway (top-down vs. bottom-up). I will change it.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -312,21 +326,97 @@ public void run(PrintStream out) throws Exception {
...
+            request.getExpressions().stream().forEach( exp -> expressions.add(exp.getName()) );
+        }
+        
+        CqlToElmTranslator cqlTranslator = getCqlTranslator();
+        CqlLibraryProvider libraryProvider = createLibraryProvider();
+        DataTypeRequirementsProcessor requirementsProcessor = new DataTypeRequirementsProcessor(cqlTranslator);
+        
+        Map<String,Set<StringMatcher>> pathsByDataType = new HashMap<>();
+        for( Map.Entry<CqlLibraryDescriptor, Set<String>> entry : expressionsByLibrary.entrySet() ) {
+            LOG.debug("Extracting data requirements for {}", entry.getKey().toString());
```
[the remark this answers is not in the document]
csandersdev: Agreed. Will change.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/optimizer/PathCaptureContext.java
```diff
@@ -0,0 +1,119 @@
...
+        return pathsByQName;
+    }
+    
+    private Map<String, VersionedIdentifier> modelIdByUri = new HashMap<>();
+    public Map<String,VersionedIdentifier> getModels() {
+        return modelIdByUri;
+    }
+    
+    public void reportProperty(Property elm) {
+        LOG.trace("Property {} source {}", elm.getPath(), elm.getSource() != null ? elm.getSource().getClass().getSimpleName() : null);
```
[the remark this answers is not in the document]
csandersdev: Will add a log guard.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/optimizer/PathCaptureContext.java
```diff
@@ -0,0 +1,119 @@
...
+            }
+            modelTypeNames = ElmUtils.getModelTypeNames(letContext.getLetClause().getExpression());
+        } else {
+            // There are times when the scope is null. I've noticed this particularly when referencing properties
+            // of another expression result
+            modelTypeNames = ElmUtils.getModelTypeNames(elm.getSource());
+        }
+        LOG.trace("ModelTypeNames {}", modelTypeNames);
+        
+        if( modelTypeNames != null ) {
```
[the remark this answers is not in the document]
csandersdev: I suppose that is true now even if it wasn't always true during dev. I think it is a reasonably defensive check though and am not inclined to change it.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/optimizer/PathCaptureVisitor.java
csandersdev: I found the expression and property trace (see PathCaptureContext) to be the most useful during debugging, but I can see how it might be helpful in the future to have additional detail. I can update the trace logging to be more consistent.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/optimizer/QueryContext.java
csandersdev: Let contexts are weird. You _can_ actually reference them before they exit. There is some special logic in place to deal with that issue.

dkwasny-ibm (approved):

msargentibm (approved):
Did a live walkthrough with Corey and then skimmed the PR. Looks good to me

PR #189, opened 2021-10-19
Add support for reading config files from COS

msargentibm:
Here's the code to read input files (model info, context defs, jobs file, cql files) from either local disk or COS. Previously the code assumed these files would always be a local read.

My intention was to add a second fvt test for Spark that would run the existing test case again, but instead of passing in local paths to each input file the test would use s3a:// paths for the file reads. I've been having issues with the toolchain side of things for a couple days, so I figured I'd get the code in front of folks to start gathering feedback.

### DIFF
--- a/cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/library/HadoopBasedCqlLibraryProvider.java
+++ b/cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/library/HadoopBasedCqlLibraryProvider.java
@@ -1,0 +1,74 @@
+/*
+ * (C) Copyright IBM Corp. 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package com.ibm.cohort.cql.library;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.LocatedFileStatus;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.RemoteIterator;
+
+public class HadoopBasedCqlLibraryProvider implements CqlLibraryProvider {
+
+	private Path directory;
+	private Configuration configuration;
+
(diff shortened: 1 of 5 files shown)


cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -245,6 +251,7 @@ public void run(PrintStream out) throws Exception {
         try (SparkSession spark = sparkBuilder.getOrCreate()) {
             boolean useJava8API = Boolean.valueOf(spark.conf().get("spark.sql.datetime.java8API.enabled"));
             this.typeConverter = new SparkTypeConverter(useJava8API);
+            this.hadoopConfiguration = new SerializableConfiguration(SparkHadoopUtil.get().newConfiguration(SparkContext.getOrCreate().conf()));
```
dkwasny-ibm, line 254: I think you can shorten this a bit by using the existing `SparkSession` like so.
```
this.hadoopConfiguration = new SerializableConfiguration(spark.sparkContext().hadoopConfiguration());
```

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/library/HadoopBasedCqlLibraryProvider.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.apache.commons.io.IOUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FSDataInputStream;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.LocatedFileStatus;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.RemoteIterator;
+
+public class HadoopBasedCqlLibraryProvider implements CqlLibraryProvider {
```
dkwasny-ibm, line 23: No unit tests targeting this class specifically?

dkwasny-ibm (changes requested):
Some small changes.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/library/HadoopBasedCqlLibraryProvider.java
dkwasny-ibm: Yea nothing crazy.  Just something that exercises all the code paths using local files in the project.

dkwasny-ibm (approved):

PR #192, opened 2021-10-21
Wffhcohort 528 spark terminology provider

tabishop:
This PR has the changes for allowing xml or json FHIR formatted ValueSets to be loaded from the the filesystem or S3. The ValueSetImporter tool was modified to allow saving of valueSets created using spreadsheets. I made usage of the filesystem based valueSets optional so if a -t option is not provided on the spark submit, they will use the UnsupportedTerminologyProvider() which will throw exceptions if they try to reference a valueset. Let me know if there are any questions.

### DIFF
--- a/cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProvider.java
+++ b/cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProvider.java
@@ -154,7 +154,7 @@ public class R4FileSystemFhirTerminologyProvider implements CqlTerminologyProvid
 		//strip of the urn or url portions of the id if they exist
 		if (valueSetInfo.getId().startsWith("urn:oid:")) {
 			valueSetId = valueSetInfo.getId().replace("urn:oid:", "");
-		} else if(valueSetInfo.getId().startsWith("http") || valueSetInfo.getId().startsWith("https") ) {
+		} else if (valueSetInfo.getId().startsWith("http")) {
 			valueSetId = valueSetInfo.getId().substring(valueSetInfo.getId().lastIndexOf("/")+1);
 		} else {
 			valueSetId = valueSetInfo.getId();
--- a/fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
+++ b/fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
@@ -34,6 +34,9 @@ import ca.uhn.fhir.rest.client.api.IGenericClient;
 
 public class ValueSetImporter {
 	private static final Logger logger = LoggerFactory.getLogger(ValueSetImporter.class.getName());
+	
+	private enum FileFormat {JSON, XML};
+	
 	public static final class ValueSetImporterArguments {
 		@Parameter(names = {"-m",
 				"--measure-server"}, description = "Path to JSON configuration data for the FHIR server connection that will be used to retrieve measure and library resources.")
@@ -52,7 +55,7 @@ public class ValueSetImporter {
 		String fileSystemOutputPath;
 		
 		@Parameter(names = {"-o", "--file-system-output-format"}, description = "Format to use when exporting value sets to the file system when using the -p/--file-system-output-path parameters. Valid values are JSON or XML. If not specified, the default output format will be JSON", required = false)
-		String filesystemOutputFormat = "json";
+		FileFormat filesystemOutputFormat = FileFormat.JSON;
 
 		@Parameter(description = "The list of value set spreadsheets to import", required = true)
 		List<String> spreadsheets;
@@ -65,11 +68,6 @@ public class ValueSetImporter {
 			
 			if(fileSystemOutputPath == null && measureServerConfigFile == null) {
 				throw new IllegalArgumentException("Parameters [-m, --measure-server] and [-p, --file-system-output-path] cannot both be null. Please supply a value for one of these parameters");
-			}
-			
-			//check for valid file system output format
-			if (filesystemOutputFormat != null && !(filesystemOutputFormat.equalsIgnoreCase("json") || filesystemOutputFormat.equalsIgnoreCase("xml"))){
-				throw new IllegalArgumentException("Bad parameter value. Valid values for [-o, --file-system-output-format parameter] are json or xml");
 			}
 		}
 	}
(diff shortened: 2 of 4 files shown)


cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProvider.java
```diff
@@ -0,0 +1,227 @@
...
+import ca.uhn.fhir.context.ConfigurationException;
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.parser.DataFormatException;
+
+/**
+ * This class is used to read ValueSet definitions from the local filesystem or S3 compatible source.
+ * ValueSet definitions are expected to be stored in FHIR xml or JSON format and can be generated from
+ * spreadsheets using the com.ibm.cohort.tooling.fhir.ValueSetImporter tool.
+ * 
+ * In order to use ValueSets stored on the filesystem, the following spark config parameters must be provided:
```
csandersdev, line 39: I don't think these docs are correct. It should say "in order to use ValueSets stored in an S3 compatible endpoint" and then the properties section needs to be updated. Hadoop uses the `spark.hadoop.fs.s3a` configuration properties first. If they aren't provided, it will fall back to using the common AWS_XXX environment variables. The two settings that can be provided as environment variables are `AWS_ACCESS_KEY_ID` (vs. `-Dspark.hadoop.fs.s3a.access.key`) and `AWS_SECRET_ACCESS_KEY` (vs. `-Dspark.hadoop.fs.s3a.secret.key`). AWS_ACCESS_KEY should be updated in the docs as shown above and AWS_ENDPOINT and AWS_LOCATION should be removed as those were specific to the Spike work and they are no longer used anywhere in our code.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProvider.java
```diff
@@ -0,0 +1,227 @@
...
+
+	/* (non-Javadoc)
+	 * 
+	 * Returns true if the provided code exists in the provided ValueSet
+	 * 
+	 * @see org.opencds.cqf.cql.engine.terminology.TerminologyProvider#in(org.opencds.cqf.cql.engine.runtime.Code, org.opencds.cqf.cql.engine.terminology.ValueSetInfo)
+	 */
+	@Override
+	public boolean in(Code code, ValueSetInfo valueSetInfo) {
+		LOG.debug("Entry: in() ValueSet.getId=[" + valueSetInfo.getId() + "] version=[" + valueSetInfo.getVersion() +"]");
```
csandersdev, line 76: SLF4J provides parameterized message support out of the box. You can use Java MessageFormat style syntax (e.g. {} placeholders) and then pass the arguments for those in a varargs parameter after the message. It allows for SLF4J to short circuit string creation when the logging is disabled. This is similar to why people use log guards (e.g. if( LOG.isDebugEnabled()) in some older log4j code. It is recommended to use the parameterized or guarded version for optimal performance. Instead of me tagging every line with this note, please consider this comment applicable for all log messages that come after.

cohort-evaluator-spark/pom.xml
```diff
@@ -68,6 +68,22 @@
...
+		<dependency>
+			<groupId>ca.uhn.hapi.fhir</groupId>
+			<artifactId>org.hl7.fhir.r4</artifactId>
+		</dependency>
+		
+		<dependency>
+			<groupId>ca.uhn.hapi.fhir</groupId>
+			<artifactId>hapi-fhir-structures-r4</artifactId>
+			<scope>runtime</scope>
+		</dependency>
```
csandersdev, line 87: We should have a team discussion on the use of HAPI FHIR in this location. I've been very specifically avoiding any HAPI dependencies for the Spark work. One one hand, it is consistent with our work in other areas of the project. On the other hand, I think we would like to move away from HAPI and instead use IBM FHIR for all our work. The IBM FHIR model objects are slightly harder to use, but bring a bunch of improvements in terms of data consistency and validation in addition to the eat your own dogfood benefit.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -44,9 +47,31 @@
...
+		String filesystemOutputFormat = "json";
 
 		@Parameter(description = "The list of value set set spreadsheets to import", required = true)
 		List<String> spreadsheets;
+		
+		public void validate() {
+			//check to make sure we are either exporting to file OR importing to a fhir server, but not both
+			if(fileSystemOutputPath != null && measureServerConfigFile != null) {
+				throw new IllegalArgumentException("Parameters [-m, --measure-server] and [-p, --file-system-output-path] cannot both be specified on the same invocation");
+			}
```
csandersdev, line 64: I think there is a world where running with both filesystem output and upload to FHIR server enabled makes sense. For example, if the valueset upload to FHIR was failing, I might want to be able to see the file that was being produced to check its contents after the failure occurred. The MeasureImporter does this by default. It always writes out the FHIR resources that it creates so that I can review the request and response bundles.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -73,9 +105,35 @@ static void runWithArgs(String[] args, PrintStream out) throws IOException {
...
+						String vsFileName = valueSetId + "." +arguments.filesystemOutputFormat.toLowerCase();
+						
+						//create the output dir if it doesn't exist
+						File outputDir = new File(arguments.fileSystemOutputPath);
+						if(!outputDir.exists()) {
+							outputDir.mkdir();
+						}
+						
+						//write to xml or json format
+						BufferedWriter writer = new BufferedWriter(new FileWriter(arguments.fileSystemOutputPath + System.getProperty("file.separator")+vsFileName));
```
csandersdev, line 130: This should be in a try-with-close block (e.g. try( BufferedWriter writer = new .... ) { /* do something */ }

csandersdev: On a separate note, I wonder if we should introduce a hadoop dependency here or if local filesystem is ok? The ultimate destination for the files for most consumers is going to be an S3 path.

csandersdev (changes requested):

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProvider.java
tabishop: Good suggestion will change

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
tabishop: I agree, but I think that would be an additional enhancement to the ValueSetImporter tool that was not in the scope of this work item. If we deem that enhancement necessary, I would agree that this would be changed as part of that.

PR #197, opened 2021-11-01
Cql spark evaluator asoc updates

tabishop:
Some updates to our dependency versions to address AppScan on Cloud issues found in the scan.


cohort-evaluator-spark/pom.xml
```diff
@@ -182,6 +182,12 @@
 		<dependency>
 			<groupId>org.apache.hadoop</groupId>
 			<artifactId>hadoop-common</artifactId>
+			<exclusions>
+				<exclusion>
+					<groupId>io.netty</groupId>
+					<artifactId>netty</artifactId>
+				</exclusion>
+			</exclusions>
```
csandersdev, line 190: Did you consider putting this exclusion in the parent POM? If we don't want netty with Hadoop, we probably don't want it everywhere we happen to use the hadoop dependency.

tabishop: Yes, meant for it to go in the parent. Good catch.

csandersdev (approved):
LGTM

PR #198, opened 2021-11-01
Wffhcohort 649 - Capture CQL evaluation errors

msargentibm:
These changes allow a user to:

1. Optionally prevent the Spark program from failing due to errors during CQL evaluation and write out a summary or any errors encountered.
2. Optionally output a `_SUCCESS` marker file to a configured location if the Spark program finishes execution without encountering any errors.

Using Jackson and `EvaluationSummary` might be a little overkill for this task, but my thought was that `EvaluationSummary` might grow over time and could be a container for any sort of extra summary information we wish to include from a particular run.

### DIFF
--- a/cohort-evaluator-spark/pom.xml
+++ b/cohort-evaluator-spark/pom.xml
@@ -114,6 +114,11 @@
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.fasterxml.jackson.core</groupId>
+			<artifactId>jackson-core</artifactId>
+			<scope>compile</scope>
+		</dependency>
 		
 		<dependency>
 			<groupId>xpp3</groupId>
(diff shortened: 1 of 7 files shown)


rickjstevens:
Overall, I think this change looks good. Just wanted to make sure I am understanding level of detail we would write for errors. Am I correct in assuming we would have separate entry in the error list for each "row" and CQL define that failed to evaluate for a given context - e.g. if we were evaluating a single define against a patient context and evaluation of that define failed for 4 patients, we would see 4 entries in the error list - one for each patient? So someone could look at the error list output to determine which specific patients (or other context instances) we failed evaluation for? And successful evaluation results for all other patients would show up in results file for that context?

msargentibm:
> if we were evaluating a single define against a patient context and evaluation of that define failed for 4 patients, we would see 4 entries in the error list - one for each patient? So someone could look at the error list output to determine which specific patients (or other context instances) we failed evaluation for? And successful evaluation results for all other patients would show up in results file for that context?

That's correct.

msargentibm:
Closing the PR so I can add in the features discussed during standup today. Will either re-open this PR or a new PR when those changes are ready.

PR #200, opened 2021-11-02
Wffhcohort-655

shelleyMurphy:
Adding PMD rule that calls out size > 0 calls and recommends replacing them with CollectionUtils.isEmpty

### DIFF
--- a/fhir-client-config/pom.xml
+++ b/fhir-client-config/pom.xml
@@ -71,5 +71,11 @@
 			<artifactId>mockito-core</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-collections4</artifactId>
+			<version>4.4</version>
+			<scope>compile</scope>
+		</dependency>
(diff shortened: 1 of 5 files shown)


fhir-client-config/pom.xml
```diff
@@ -71,5 +71,11 @@
 			<artifactId>mockito-core</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-collections4</artifactId>
```
csandersdev, line 76: This is a new dependency for all of our projects. Is there any need to introduce Collections v4 or can it be just the v1 API like the rest of our projects?

shelleyMurphy: Whoops, let my IDE add the dependency. Just rerunning the tests on the other dependency now, but I think that should be fine.

csandersdev (approved):
LGTM

tabishop (approved):
LGTM

jaigoradia (approved):

PR #201, opened 2021-11-05
Wffhcohort 649 

msargentibm:
Opening a new PR that encompasses the original work in https://github.com/Alvearie/quality-measure-and-cohort-service/pull/198, plus adding in extra information to the evaluation summary that is created.

The program now requires a user to pass in a metadata path. We write a `_SUCCESS` marker file in that folder if the Spark job completes without any CQL evaluation errors. We also write out a batch summary file. Here is an example file:

```
{
  "applicationId" : "spark-39c137844a6241bf99115f4ef383d58d",
  "startTimeMillis" : 1636127395548,
  "endTimeMillis" : 1636127435798,
  "totalContexts" : 1,
  "executionsPerContext" : {
    "Patient" : 10
  }
}
```

Additionally, if the `--halt-on-error` option is false when the program runs, then any CQL evaluation errors will be accumulated and reported in the batch summary file. Example:
```
{
  "applicationId" : "local-1636067021219",
  "startTimeMillis" : 1636067022136,
  "endTimeMillis" : 1636067029319,
  "totalContexts" : 1,
  "executionsPerContext" : {
    "Patient" : 100
  },
  "errorList" : [ {
    "contextName" : "Patient",
    "contextId" : "pat_51",
    "outputColumn" : "cohort",
    "exception" : "Expected a list with at most one element, but found a list with multiple elements."
  }, {
    "contextName" : "Patient",
    "contextId" : "pat_40",
    "outputColumn" : "cohort",
    "exception" : "Some other error message"
  }]
}
 ```

I verified that the application id matches the id reported in the Spark history server:
Screen Shot 2021-11-05 at 11 53 31 AM

### CODE REVIEW

**Comment by msargentibm** [comment]
I still need to rework the docs and rebase onto the feature branch this afternoon.

File: cohort-evaluator-spark/pom.xml [hunk 1]
```diff
@@ -114,6 +114,11 @@
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.fasterxml.jackson.core</groupId>
+			<artifactId>jackson-core</artifactId>
+			<scope>compile</scope>
```
csandersdev, line 120: Pedantic: compile scope is the default and thus doesn't need to be specified.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -155,6 +170,34 @@ public void testAllTypesEvaluationSuccess() throws Exception {
         validateOutputCountsAndColumns(bFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureB|cohort")), 575, "parquet");
         validateOutputCountsAndColumns(cFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureC|cohort")), 600, "parquet");
         validateOutputCountsAndColumns(dFile.toURI().toString(), new HashSet<>(Arrays.asList("id", "MeasureD|cohort")), 567, "parquet");
+        
+        assertTrue(new File(metadataDir, HadoopPathOutputMetadataWriter.SUCCESS_MARKER).exists());
+
+		Set<Path> summaryFilesAfter = getSummaryFilesInPath(metadataDir.toPath());
+		
+		summaryFilesAfter.removeAll(summaryFilesBefore);
+		assertEquals(1, summaryFilesAfter.size());
```
csandersdev, line 179: There is whitespace inconsistency through here

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorTest.java
```diff
@@ -125,8 +135,12 @@ public void testAllTypesEvaluationSuccess() throws Exception {
         File bFile = new File(outputDir, "B_cohort");
         File cFile = new File(outputDir, "C_cohort");
         File dFile = new File(outputDir, "D_cohort");
+        
+        File metadataDir = new File(outputDir, "evaluation_success");
 
-        String [] args = new String[] {
+		Set<Path> summaryFilesBefore = getSummaryFilesInPath(metadataDir.toPath());
+
+		String [] args = new String[] {
```
csandersdev, line 143: Whitespace inconsistency

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/metadata/HadoopPathOutputMetadataWriter.java
```diff
@@ -0,0 +1,54 @@
...
+    
+    @Override
+    protected void writeBatchSummary(EvaluationSummary evaluationSummary) {
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectWriter writer = mapper.writer(new DefaultPrettyPrinter());
+        try (FSDataOutputStream outputStream = metadataPath.getFileSystem(hadoopConfig).create(metadataPath.suffix("/" + BATCH_SUMMARY_PREFIX + evaluationSummary.getApplicationId()))) {
+            String jsonString = writer.writeValueAsString(evaluationSummary);
+            outputStream.write(jsonString.getBytes());
+        } catch (IOException e) {
+			throw new RuntimeException("Error writing batch summary report", e);
```
csandersdev, line 51: whitespace inconsistency

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/metadata/BaseOutputMetadataWriter.java
```diff
@@ -0,0 +1,29 @@
...
+ * (C) Copyright IBM Corp. 2021
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.ibm.cohort.cql.spark.metadata;
+
+
+import org.apache.commons.collections.CollectionUtils;
+
+public abstract class BaseOutputMetadataWriter {
```
csandersdev, line 11: I was expecting to find an OutputMetadataWriter interface implemented here. Did you decide against that?

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -300,17 +311,36 @@ public void run(PrintStream out) throws Exception {
...
+            if (args.metadataOutputPath != null) {
+                Path metadataPath = new Path(args.metadataOutputPath);
+                
+                if (errorAccumulator != null) {
+                    evaluationSummary.setErrorList(errorAccumulator.value());
+                }
+                
+                evaluationSummary.setTotalContexts(contextAccum.value());
+                
+                HadoopPathOutputMetadataWriter writer = new HadoopPathOutputMetadataWriter(metadataPath, hadoopConfiguration.value());
```
csandersdev, line 338: The LHS of this expression should be OutputMetadataWriter and I'd like to see the RHS be a function call on this class `getOutputMetadataWriter()` so that people can override the behavior if they want.

cohort-evaluator-spark/pom.xml
```diff
@@ -178,6 +183,12 @@
 			<groupId>commons-io</groupId>
 			<artifactId>commons-io</artifactId>
 		</dependency>
+		
+		<dependency>
+			<groupId>commons-collections</groupId>
+			<artifactId>commons-collections</artifactId>
+			<scope>compile</scope>
```
csandersdev, line 190: Same comment as above about redundant scope

csandersdev (commented):
I left a few minor comments. Not approving yet as you pointed out the docs still need to be updated.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/metadata/BaseOutputMetadataWriter.java
msargentibm: I still couldn't figure out what made sense to  have in the interface.  Do you have a concrete idea?

msargentibm: I added an interface that has `void writeMetadata(EvaluationSummary evaluationSummary);`. Wasn't sure if you had more than that in mind.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -538,28 +574,44 @@ protected ContextDefinitions readContextDefinitions(String path) throws Exceptio
...
+                        String outputColumnKey = columnEncoder.getColumnName(request, entry.getKey());
+                        expressionResults.put(outputColumnKey, typeConverter.toSparkType(entry.getValue()));
+                    }
+                } catch (Throwable th) {
+                    if (errorAccum != null) {
+                        Object contextId = rowsByContext._1();
+                        errorAccum.add(new EvaluationError(contextName, contextId, singleRequest.getExpressionNames().iterator().next(), th.getMessage()));
+                    }
+                    else {
+                        throw new RuntimeException("CQL evaluation failed: ", th);
```
dkwasny-ibm, line 612: The new exception is not as verbose as the original.
Any particular reason to simplify the error message?

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/metadata/BaseOutputMetadataWriter.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package com.ibm.cohort.cql.spark.metadata;
+
+
+import org.apache.commons.collections.CollectionUtils;
+
+public abstract class BaseOutputMetadataWriter implements OutputMetadataWriter{
+    
+    public static String SUCCESS_MARKER = "_SUCCESS";
+    public static String BATCH_SUMMARY_PREFIX = "batch_summary-";
```
dkwasny-ibm, line 14: Gimme a little `final` here and I'll shut up 😄

dkwasny-ibm (changes requested):
Just some small requests.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
msargentibm: We don't want context keys making their way to LogDNA.  I suppose I could still report the context name and the request so there is some additional info.

dkwasny-ibm (approved):

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -538,28 +574,46 @@ protected ContextDefinitions readContextDefinitions(String path) throws Exceptio
...
+                        String outputColumnKey = columnEncoder.getColumnName(request, entry.getKey());
+                        expressionResults.put(outputColumnKey, typeConverter.toSparkType(entry.getValue()));
+                    }
+                } catch (Throwable th) {
+                    if (errorAccum != null) {
+                        Object contextId = rowsByContext._1();
+                        errorAccum.add(new EvaluationError(contextName, contextId, singleRequest.getExpressionNames().iterator().next(), th.getMessage()));
+                    }
+                    else {
+                        throw new RuntimeException(String.format("CQL evaluation failed for ContextName: %s, Request: %s",
```
msargentibm, line 612: @csandersdev , do you have any thoughts on if the CqlEvaluationRequest itself is ok to include in the message here? I know it has a `conextValue`, but it seems like that's always set to `NA` right now and we don't  do anything  with it. I wasn't sure if this could be used in the future and could ever maybe be set to a context id we wouldn't want to log.

jaigoradia (approved):

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
csandersdev: Right now, `contextValue` is always user-populated based on what is in the jobs file and we always set that to NA. I think there is a reasonable argument that we should populate that during evaluation with the actual context value, but that isn't something we do today and it wouldn't impact runtime behavior since we group the data ahead of time. The only reason `contextValue` exists in the jobs file at all is because I was trying to model a request object that could be used also in the REST-FHIR path (you'll notice the object lives in cohort-evaluator not cohort-evaluator-spark). It certain creates extra noise in the jobs file and could maybe be dropped. When we added the output column name, it already started to get more Spark-specific than general, so I'm not quite sure what it should be right now.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/metadata/BaseOutputMetadataWriter.java
csandersdev: I am satisfied with that.

csandersdev (approved):
LGTM

PR #204, opened 2021-11-10
Wffhcohort 702 -- Add fields to batch summary for performance work

msargentibm:
For the performance work, we would like to have the total runtime and runtime per context available in the batch summary report that gets created. This PR adds those fields to the report and sets the values from the main program.


jaigoradia (approved):
LGTM

dkwasny-ibm (approved):

PR #205, opened 2021-11-12
Index codes in valuesets for faster lookups

tabishop:
This PR changes our file system based TerminologyProvider to cache sets of codes instead of the fhir valueSet objects themselves. I implemented a simple wrapper class around the Code object which provides hashCode and equals functions that work for our terminology implementation. Using sets should improve the lookup time of the codes within the valuesets especially for large valuesets.

Signed-off-by: TimBishop <tabishop@us.ibm.com>


cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/HashedCode.java
```diff
@@ -0,0 +1,60 @@
...
+	public Code getCodeObject() {
+		return code;
+	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		//only hash the code value since it is possible to get codes passed with
+		//without the other fields
```
jillrdoty, line 51: I believe that only hashing the code would allow for collisions if the same code existed in 2 different systems or with 2 different versions.  

If the value set contained ICD 250 v1 and ICD 250 v2, when loaded only one of these would be represented in the hashset.  If the comparison code only has the code we're fine, but it it also has the version and it's looking for v1 when only v2 is represented I think we have a problem.

I haven't written any test cases to prove this happens so happy if you can explain I'm wrong.

cohort-evaluator-spark/Dockerfile
```diff
@@ -53,7 +53,7 @@ ENV SPARK_HOME=${SPARK_HOME}
 # Become root for the Spark installation process
 USER root
 # Perform a full update of all packages to minimize vulnerabilities
-RUN dnf update -y &&\
+RUN dnf update --nobest -y &&\
```
msargentibm, line 56: Is `--nobest` meant to be here permanently? Or is this just so you could do a test run?

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/HashedCode.java
```diff
@@ -0,0 +1,60 @@
...
+
+
+/**
+ * This class is a simple wrapper class providing equals and hashCode functions
+ * for the org.opencds.cqf.cql.engine.runtime.Code class
+ * This is used to enable Set lookups for codes within ValueSets, which will be
+ * more performant than iterating through a list of codes
+ *
+ */
+public class HashedCode {
```
csandersdev, line 20: This looks to be a duplicate of the CodeKey object in cohort-model-datarow. Do we need both?

msargentibm, line 20: Did you look at our `CodeKey` class at all? It only uses the code and system for comparisons, but I wasn't sure if that's sufficient for the indexing you are hoping to add.

tabishop: I did look at the CodeKey object, but in org.opencds.cqf.cql.engine.elm.execution.InValueSetEvaluator.inValueSet(), around line 49, there is a branch that passes in a code object with only a code specified and no system. Also, in the R4RestFhirTerminologyProviderTest, there is a testcase inOperationHandlesNullSystem() which implies that we need to support compare checks against code with no system defined.

For this reason, I needed to create my own class, because CodeKey does equality checks where the code system must be either present and equal on both codes being compared, or null on both. It doesn't support it being set on the code in the ValueSet read in from the file system, but null on the one passed into the in() method.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/HashedCode.java
tabishop: There will be collisions but java handles collisions of 2 objects with the same hashCode by falling back to the equals() method. So objects with the same hashcode basically end up in a list and iterated over. I did add some tests to test against codes with different code system versions to verify, and did find a bug in that the valueSetImporter was not reading in the code system version from the spreadsheet, so we have a bug to fix here that I will address before I deliver.

Also, see my note below to Corey's comment on usage of the CodeKey class and why I implemented the equals method the way I did.

cohort-evaluator-spark/Dockerfile
tabishop: This is temporary for now to allow testing, but may become permanent pending outcome of discussions with Frances.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/HashedCode.java
tabishop: See my note above to Corey on this topic.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/HashedCode.java
jillrdoty: Thanks!

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/HashedCode.java
msargentibm: > For this reason, I needed to create my own class

It's definitely one option. Another option would be to use `CodeKey` and build up an index explicitly using a Map or something rather than obscuring the logic in the `equals` and `hashCode` methods.

msargentibm:
I was looking at the Terminology Operators section in the CQL Author's Guide: https://cql.hl7.org/02-authorsguide.html#terminology-operators.

The note there mentions:
> Note that valueset membership is based strictly on the definition of equivalence (i.e. two codes are the same if they have the same value for the code and system elements, ignoring display and version). 

Given this note, it seems like the logic that allows a code with no system to match a code with a system conflicts with the CQL spec.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/HashedCode.java
csandersdev: I spent a lot of time on this today. I don't think there is clarity between the spec and the FHIR reference implementation on how this should work. The FHIR R4 Terminology Provider behaves in a way that isn't actually supported by FHIR (e.g. not providing the system parameter in the $validate-code request). The Javascript engine behaves in a clear way and reasonable way, but I don't think it is strictly matched up with the spec. I wrote up my notes on this topic and posted the question in [Zulip](https://chat.fhir.org/#narrow/stream/179220-cql/topic/'String'.20in.20MyValueSet) to see what the community has to say about it.

csandersdev: Per Bryn, the correct behavior is documented in the CQL reference section [here](https://cql.hl7.org/09-b-cqlreference.html#in-valueset). 

> For the String overload, if the given valueset contains a code with an equivalent code element, the result is true. Note that for this overload, because the code being tested cannot specify code system information, if the resolved value set contains codes from multiple code systems, a run-time error is thrown because the operation is ambiguous.

I think the code index should be Map<String,Set<String>> where the key is the code and the value is the set of distinct codesystems associated with that code. When the lookup occurs on a code with no system, the provider should throw an "ambiguous lookup" exception.

```java
Set<String> systems = codeIndex.get(code);
if( systems != null && systems.size() > 1) { 
  throw new IllegalArgumentException("Ambiguous code lookup of code foo under valueset bar"); 
}
```

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProviderTest.java
```diff
@@ -29,12 +29,17 @@
 import org.opencds.cqf.cql.engine.terminology.CodeSystemInfo;
 import org.opencds.cqf.cql.engine.terminology.ValueSetInfo;
 
+import com.ibm.icu.impl.Assert;
```
csandersdev, line 32: This is a mistaken auto-import. We shouldn't be using any com.ibm.icu classes. You probably meant to grab `org.junit.Assert`

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProvider.java
```diff
@@ -210,8 +197,26 @@ public boolean accept(Path path) {
 								.parseResource(new InputStreamReader(fileSystem.open(valueSetFiles[0].getPath())));
 						LOG.info("Unmarshalled json {}", valueSetFhirR4.getId());
 					}
+					
+					//Add the codes to a HashSet using a wrapper object with the correct hashCode/equals behavior
```
csandersdev, line 201: Your comment here is a little out of date now since you are no longer using the custom Code object with hashCode/equals logic.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProviderTest.java
tabishop: Good catch, was wondering why the build kept barfing on that.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProvider.java
tabishop: fixed

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/terminology/R4FileSystemFhirTerminologyProviderTest.java
```diff
@@ -112,12 +199,24 @@ public void inOperationHandlesNullSystem() throws Exception {
...
+	@Test(expected = IllegalArgumentException.class)
+	public void inOperationHandlesMultipleCodeSystemCode() throws Exception {
+		ValueSetInfo info = new ValueSetInfo();
+		info.setId(setId);
+
+		Code code = new Code();
+		code.setCode(TEST_CODE_MULTIPLE_CODE_SYSTEMS);
+
+		boolean result = provider.in(code, info);
+		assertTrue(result);
```
msargentibm, line 212: Do you need this assertion still with the expected exception already a part of the test?

tabishop: removed, although I don't think it makes much difference

csandersdev (changes requested):
I gave it one last once over and found a small issue in the ambiguity checking logic.

csandersdev (approved):
LGTM

PR #208, opened 2021-11-18
WFFHCOHORT-689 - Split output into multiple rows based on CQL parameters

csandersdev:
Instead of every evaluation result per context being joined into one
large row, consumers have asked that we split related evaluation results
into separate rows. For example, if the same logic is evaluated for
multiple time ranges, the output will contain a column for each piece of
logic and a row for each time range.

Rows are split based on a grouping key. The grouping key consists of one
or more CQL parameter values. By default all parameter values are used.
In the case where there are more CQL parameter values than are needed
for the grouping key, a new program argument `--key-parameter` is
provided. This parameter can be repeated as many times as needed to
define the exact set of parameters that define the grouping key. The
behavior can be disabled by using the `--disable-result-grouping`
argument.


cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -730,28 +771,16 @@ protected CqlEvaluationRequests readJobSpecification(String path) throws Excepti
...
-                        .map(RowFactory::create),
+                resultsByContext.values(),
                 schema
         );
 
-		if (args.outputPartitions != null) {
-			dataFrame = dataFrame.repartition(args.outputPartitions);
-		}
+        if (args.outputPartitions != null) {
+        	dataFrame = dataFrame.repartition(args.outputPartitions);
```
dkwasny-ibm, line 782: I think the 3rd leading tab is still here.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -593,8 +614,13 @@ protected ContextDefinitions readContextDefinitions(String path) throws Exceptio
         
         List<CqlEvaluationRequest> requestsForContext = requests.getEvaluationsForContext(contextName);
         
-        Map<String, Object> expressionResults = new HashMap<>();
+        // parameters json -> {columnName, result}
+        Map<String,Map<String, Object>> expressionResultsByParameters = new HashMap<>();
         for (CqlEvaluationRequest request : requestsForContext) {
+            
+            String parametersJson = CqlEvaluationRequestUtil.getKeyParametersColumnData(request, getKeyParameterNames());
```
dkwasny-ibm, line 621: I'm not a big fan of creating a JSON string from a POJO for every request for every context value.
If I'm not mistaken, the resulting JSON will not differ between context values.
I'd like to see some way of pre-computing the strings once in the driver and sent to the executors either via serialization (i.e. as a property in `SparkCqlEvaluator`) or broadcast variables.
Caching the serialized form in a `transient` variable somewhere would work as well.

If you can confidently say that this results in a negligible performance loss overall, then I'm OK with leaving it.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorArgs.java
```diff
@@ -91,4 +91,10 @@
     
     @Parameter(names = { "--debug" }, description = "Enables CQL debug logging")
     public boolean debug = false;
+
+    @Parameter(names = { "--disable-result-grouping" }, description = "Disable use of CQL parameters to group context results into separate rows", required = false)
+    public boolean disableResultGrouping = false;
+    
+    @Parameter(names = { "--key-parameter" }, description = "One or more parameter names that should be included in the parameters column for output rows that are generated.", required = false)
```
dkwasny-ibm, line 98: Should this be called `--key-parameters`?

dkwasny-ibm (commented):
Some rough thoughts.
The only real concern is the extra work spent serializing JSON objects.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorArgs.java
csandersdev: I considered it, but the value of each parameter is a single string. You would specify multiple args to get multiple values. For example, `--key-parameter EndDate --key-parameter MinimumAge` reads better to me than `--key-parameters EndDate --key-parameters MinimumAge`. JCommander doesn't support list parsing out of the box for something like `--key-parameters EndDate,MinimumAge`. We could implement a custom arg parser for that ourselves if we wanted, but I've been just sticking to the core functionality.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
csandersdev: Will fix.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
csandersdev: I added a cache for the calculated parameters JSON.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorArgs.java
csandersdev: I got curious about this today since wanting a simple comma-separated list of values comes up a lot. It turns out that JCommander _does_ support this natively, but only for parameters that are List types. It doesn't work for Set parameters even if you implement a custom listConverter class. I changed the Set parameters to List parameters and updated the code to support the new type. I also updated the long names of some of the existing parameters to reflect their multi-valued nature and for consistency across parameters.

msargentibm (commented):
Did my first pass and only had a README comment so far. I pulled the branch down and poked around a bit manually and none of my manual tests broke anything.

I plan on taking another look tomorrow and will approve if I don't come across anything strange.

docs/user-guide/spark-user-guide.md
```diff
@@ -50,43 +54,46 @@ Usage: SparkCqlEvaluator [options]
...
       Print help text
     --input-format
-      Spark SQL format identifier for input files. If not provided, the value
+      Spark SQL format identifier for input files. If not provided, the value 
       of spark.sql.datasources.default is used.
   * -i, --input-path
-      Key-value pair of resource=URI controlling where Spark should read
-      resources referenced in the context definitions file will be read from.
+      Key-value pair of resource=URI controlling where Spark should read 
+      resources referenced in the context definitions file will be read from.
```
jaigoradia, line 61: i think this should say 'Key-value pair of .... where Spark should read resources referenced in the context definitions file from.   "will be read from"  is redundant.

docs/user-guide/spark-user-guide.md
```diff
@@ -326,9 +333,13 @@ If the user wishes to limit which jobs are evaluated during a specific applicati
...
+
+The parameters column contains JSON data representing the CQL parameters that were used in the evaluation of the column results. The format of the JSON data is the same as the parameter data in the jobs file. This column allows the results of a single context evaluation to be split into multiple rows of related columns. For example, an input parameter might be `EndDate` for the logic and the desired result is for there to be a separate record in the output for each CQL context + end date.
+
+Row grouping by parameters is enabled by default or can be disabled using the `--disable-row-grouping` flag. When enabled, all parameter values are used as the grouping key. In the case where there are more CQL parameter values than are needed for the grouping key, the `--key-parameters` program argument can be used to limit which parameter values are used as the grouping key. Multiple values can be provided in a comma-separate list or this parameter can be repeated as many times as needed to define the exact set of parameters needed.
```
jaigoradia, line 340: The first sentence should probably be split into two.  Like so:  Row grouping by parameters is enabled by default. It can be disabled by using the `--disable-row-grouping` flag.  I feel this little section could be made more understandable by including  a simple example of what the output looks like when CQL parameters are used.

msargentibm (approved):
Forgot to approve last week. LGTM

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/util/CqlEvaluationRequestUtil.java
```diff
@@ -0,0 +1,44 @@
...
+            
+            Map<String,Parameter> keyParameters = req.getParameters();
+            if( keyParameters != null ) {
+                if( keyParameterNames != null ) {
+                    keyParameters = keyParameters.entrySet().stream()
+                            .filter( e -> keyParameterNames.contains(e.getKey()) )
+                            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+                }
+                try {
+                    parametersJson = new ObjectMapper().writeValueAsString(keyParameters);
```
deekim, line 36: I think the filtering aspect of this cache setup makes it a bit more confusing
but narrowing the cache `Map` to be keyed on `keyParameters` (instead of the entire `CqlEvaluationRequest`) might be a bit cleaner?

deekim (approved):
LGTM.
My comment in `CqlEvaluationRequestUtil` is non-blocking.

jaigoradia (approved):
LGTM other than my comments related to the Readme (i.e. documentation).

shelleyMurphy (approved):
👍

docs/user-guide/spark-user-guide.md
csandersdev: I left the first sentence intact as it read pretty clearly to me and I didn't like how choppy it felt split into two. I did, however, break out the parameters stuff into its own subheading and I added some examples.

jaigoradia: Looks good.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/util/CqlEvaluationRequestUtil.java
csandersdev: I agree there was some code smell here. The cache key did not match up with the parameters declaration. I updated this class so that the key parameters are now an immutable part of the class state and only the request is passed into the getKeyParametersJson method. I also added the functionality related to the rowGroupingDisabled flag. Now, all of the parameters-related logic is in a single place which should address another comment I had received in a separate, offline conversation.

tabishop (approved):
Not my immediate area of expertise, but I didn't see any obvious errors.

dkwasny-ibm (approved):
Signed-off-by: David Kwasny 👍

PR #209, opened 2021-11-23
Sonar bug fixes

jillrdoty:
Signed-off-by: Jill Doty <jdoty@us.ibm.com>

See each change for explanation.


cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkDataRow.java
```diff
@@ -99,7 +99,7 @@ protected Object doDefaultConversion(Object sparkVal) {
         try {
             result = typeConverter.toCqlType(sparkVal);
         } catch (IllegalArgumentException ex) {
-            throw new IllegalArgumentException(String.format("Failed to convert field '%s': %s", ex.getMessage()), ex);
+            throw new IllegalArgumentException(String.format("Failed to convert field '%s': %s", sparkVal, ex.getMessage()), ex);
```
jillrdoty, line 102: Missing argument.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/data/SparkTypeConverter.java
```diff
@@ -206,7 +206,7 @@ public Object toSparkDecimal(Object obj) {
             if( obj instanceof BigDecimal ) {
                 result = (BigDecimal) obj;
             } else if( obj instanceof Number ) {
-                result = new BigDecimal(((Number) obj).doubleValue());
+                result = BigDecimal.valueOf(((Number)obj).doubleValue());
```
jillrdoty, line 209: Because of floating point imprecision, you’re unlikely to get the value you expect from the BigDecimal(double) constructor.

From the JavaDocs:

    The results of this constructor can be somewhat unpredictable. One might assume that writing new BigDecimal(0.1) in Java creates a BigDecimal which is exactly equal to 0.1 (an unscaled value of 1, with a scale of 1), but it is actually equal to 0.1000000000000000055511151231257827021181583404541015625. This is because 0.1 cannot be represented exactly as a double (or, for that matter, as a binary fraction of any finite length). Thus, the value that is being passed in to the constructor is not exactly equal to 0.1, appearances notwithstanding.

Instead, you should use BigDecimal.valueOf, which uses a string under the covers to eliminate floating point rounding errors, or the constructor that takes a String argument.

cohort-evaluator-spark/src/test/java/com/ibm/cohort/cql/spark/data/SparkDataRowTest.java
```diff
@@ -153,8 +153,8 @@ public void testConversionSemanticsNullValues() {
             assertEquals( pojo.getShortField(), sdr.getValue("shortField") );
             assertEquals( pojo.getIntegerField(), sdr.getValue("integerField") );
             assertEquals( pojo.getLongField(), sdr.getValue("longField") );
-            assertEquals( pojo.getFloatField(), (BigDecimal) sdr.getValue("floatField") );
```
jillrdoty, line 156: Comparing incompatible types should fail (it did not because this is the null test).  

I recognize that in other tests we're converting the float and double fields to BigDecimals, so if the preference is to cast pojo.getFloatField() to a BigDecimal instead of removing it from the right side I am willing to change.

cohort-evaluator/src/main/java/com/ibm/cohort/cql/evaluation/CqlContextFactory.java
```diff
@@ -66,16 +66,20 @@ public ContextCacheKey(
 
         @Override
         public boolean equals(Object o2) {
-            boolean isEqual = o2 != null && o2 instanceof ContextCacheKey;
```
jillrdoty, line 69: The instanceof check implies non-null.

The equals check will fail even if both "this" and o2 are null, but as that is the current behavior I choose not to alter that.  Happy to change if desired.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/optimizer/GraphWalkingElmVisitor.java
```diff
@@ -82,41 +82,42 @@ public R visitLibrary(Library elm, C context) {
...
             visited.add(ref);
             
             Library library = prepareLibraryVisit(getCurrentLibraryIdentifier(), ref.getLibraryName(), context);
             try {
                 // The TranslatedLibrary class that the CqlTranslator produces has this already indexed. If we want to 
                 // use any ELM without relying on the translator, we need to do the lookup ourselves. It is worth
                 // considering whether or not we build our own indexing helper instead.
-                ExpressionDef ed = library.getStatements().getDef().stream().filter( def -> def.getName().equals(ref.getName()) ).findAny().get();
-                result = visitElement(ed,context);
+            	Optional<ExpressionDef> optionalEd = library.getStatements().getDef().stream().filter( def -> def.getName().equals(ref.getName()) ).findAny();
```
jillrdoty, line 113: This is the change sonar requested - checking that the optional value is present before calling .get().  When I was in this class I noticed these two methods were almost identical so I consolidated the logic.

deekim (approved):

msargentibm (approved):
I don't have a strong about removing vs. moving  the `(BigDecimal)` cast in  the unit tests. Removing it seems fine to me.

cohort-evaluator/src/main/java/com/ibm/cohort/cql/evaluation/CqlContextFactory.java
dkwasny-ibm: This method really can't handle when `this` is null since it's an instance method.
If we're in an instance method, then `this` is guaranteed to be non-null.

dkwasny-ibm (approved):

shelleyMurphy (approved):

PR #213, opened 2021-12-07
Spark image promotion script updates

tabishop:
Additions to the prePromotion.sh script to allow the ability to push our spark docker image to the wh-common-rns (or other container registry). Script checks for presence of the env var required to contain the image name, pulls the image, and then pushes it to the configured container registry. I didn't do a ton of work to make some of the region, cloud url, etc. configurable because this is a short term fix until the 3.6.0 toolchain support is released, at which point, this code can be removed.


jaigoradia (approved):
LGTM

csandersdev (approved):

msargentibm (approved):

PR #214, opened 2021-12-08
Wffhcohort-738: Add correlation id to logging statements

shelleyMurphy:
Note: There are some spark log classes that don't accept the mdc property, such as TaskSetManager or certain scheduler logs. These logs don't look like they'd be particularly vital for everyday use (for example, `Registering RDD 5 (toString at String.java:2994) as input to shuffle 0` ), but I think it's important to keep in mind.

### DIFF
--- a/cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
+++ b/cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
@@ -290,7 +290,7 @@ public class SparkCqlEvaluator implements Serializable {
         
         SparkSession.Builder sparkBuilder = SparkSession.builder();
         try (SparkSession spark = sparkBuilder.getOrCreate()) {
-            spark.sparkContext().setLocalProperty("mdc." + CORRELATION_ID, MDC.get("CorrelationId"));
+            spark.sparkContext().setLocalProperty("mdc." + CORRELATION_ID, MDC.get(CORRELATION_ID));
             boolean useJava8API = Boolean.valueOf(spark.conf().get("spark.sql.datetime.java8API.enabled"));
             this.typeConverter = new SparkTypeConverter(useJava8API);
             this.hadoopConfiguration = new SerializableConfiguration(spark.sparkContext().hadoopConfiguration());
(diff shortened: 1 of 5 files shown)


msargentibm:
I know changing the batch summary file name was out of scope, but I did see this on the ticket:
> This correlation ID will be written with any log messages created by the application and also to the batch summary file that is created as output.

Looks like adding the correlation ID to the batch summary wasn't included in the acceptance criteria. Should we still plan on adding it here?

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorArgs.java
```diff
@@ -95,4 +95,7 @@
     
     @Parameter(names = { "--key-parameters" }, description = "One or more parameter names that should be included in the parameters column for output rows that are generated.", required = false)
     public List<String> keyParameterNames = null;
+
+    @Parameter(names = { "--correlation-id" }, description = "Identifier consistent between all logs of a single job", required = false)
```
tabishop, line 99: I would tweak this description to be a little more clear to users who are not already familiar with what this does. Maybe something like:  "This correlation ID will be written with any log messages created by the application and also to the batch summary file that is created as output." -> Obviously the batch summary part would be dependent on if we add that, per Matt's comment.

tabishop (changes requested):
Would just like one small change to the arg description.

shelleyMurphy:
@msargentibm I didn't include it because it was explicitly listed as out of scope. I'm assuming there was a reason for that, since the scope was added later, but I can add it if we think it's important.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluatorArgs.java
shelleyMurphy: Changed!

msargentibm:
@shelleyMurphy Gotcha. I was reading that as changing the file name was out of scope, but adding the correlation id to the file wasn't called out one way or the other in the "out of scope" and "acceptance criteria" sections.

shelleyMurphy:
@msargentibm ahh I see what you mean now, I think that's fair. I'll add that now

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/SparkCqlEvaluator.java
```diff
@@ -840,11 +843,20 @@ public static void main(String[] args) throws Exception {
...
+
+            SparkCqlEvaluator evaluator = new SparkCqlEvaluator(programArgs);
+            if (programArgs.help) {
+                commander.usage();
+            } else {
+                evaluator.run(System.out);
+            }
+        }
+        finally {
+            MDC.remove("CorrelationId");
```
dkwasny-ibm, line 860: This should reference the `static` variable right?

msargentibm (approved):
👍

dkwasny-ibm (changes requested):
Just one small change I'd like to see.

dkwasny-ibm (approved):

jaigoradia (approved):

tabishop (approved):
LGTM

PR #218, opened 2021-12-09
remove sonar, update version

tabishop:
Removing the spark sonar scan, and updating the version numbers in main branch to 2.0.1

Signed-off-by: TimBishop <tabishop@us.ibm.com>


Dockerfile
```diff
@@ -39,7 +39,7 @@ LABEL maintainer="IBM Quality Measure and Cohort Service Team" \
       description="Quality Measure and Cohort Service" \
       name="cohorting-app" \
       vendor="Alvearie Open Source by IBM" \
-      version="1.0" \
+      version="2.0.1" \
       release="1" \
```
csandersdev, line 43: Should this say release="2"? I'm not really clear on how that field works, but it feels like that is the major number from the semantic version. Or, do we even need that release tag? The [IBM FHIR team](https://github.com/IBM/FHIR/blob/1e8def81b69261b77ba175e3ec5e247a0b2a6cec/fhir-install/Dockerfile#L31) doesn't use it. This also begs the question about whether or not we need these tags in our other Dockerfiles like cohort-evaluator-spark/Dockerfile and docker/spark-history-server/Dockerfile.

Dockerfile
```diff
@@ -39,7 +39,7 @@ LABEL maintainer="IBM Quality Measure and Cohort Service Team" \
       description="Quality Measure and Cohort Service" \
       name="cohorting-app" \
       vendor="Alvearie Open Source by IBM" \
-      version="1.0" \
+      version="2.0.1" \
```
csandersdev, line 42: Is this set to 2.0.1 because it is what is in development for release after we post 2.0.0 (e.g. you will set the version to 2.0.0 on the release branch)? My first thought when looking at it was that we were skipping 2.0.0 and going directly to 2.0.1 which was weird.

tabishop: Correct. When I cut the release branch, the version for that will be 2.0.0 in the release branch. I wanted to remove the sonar file in both branches so just did the version update here to avoid having another PR.

Dockerfile
tabishop: These tags are optional in the dockerfiles, but I think are used if we want to create certified Red Hat images, which we don't have a requirement to do. Last release we didn't even update it at all. If someone tries to get the information on the docker image, I think is the only place this info shows up. I can change to release 2

dkwasny-ibm (approved):

csandersdev (approved):
LGTM

PR #221, opened 2021-12-10
updated the name of the cohort-evaluator-spark snapshot jar name with new release version number of 2.0.1

jaigoradia:
This PR is to fix an issue uncovered in the run-fvttests.sh script post merge of the feature branch into main.  The script has hard-coded spark-submit commands to run a couple of sets of spark tests as part of the cd toolchain.  Both of those commands previously referenced the older version of the snapshot jar file as in :  local:///opt/spark/jars/cohort-evaluator-spark-1.0.2-SNAPSHOT.jar and the tests started failing as spark executors could no longer be created with the reference to the old jar file.

This PR updates run-fvttests.sh script and specifically the spark-submit command strings to reference the new cohort-evaluator-spark-2.0.1-SNAPSHOT.jar

Jira workItem WFFHCOHORT-752 has been created to track the issue to fix this in a more generic fashion for the future. 


msargentibm (approved):

csandersdev (approved):
Approving with the understanding that we will revisit this for a more dynamic approach in the future.

shelleyMurphy (approved):

PR #224, opened 2021-12-21
Dynamically create Spark jar path

msargentibm:
~Getting the jar version in the CD toolchain was a bit tricky. I opted to have part of the CI toolchain dump the jar version out to a file, which is then committed to the umbrella repo. The CD toolchain can then `cat` the file to get the version that is used for the Spark jar.~

Edit: Changed the jar name in the Dockerfile instead.


csandersdev (approved):
LGTM.

tabishop (commented):
I'm wondering if a better approach would be to rename the cohort-evaluator-spark-<version>-SNAPSHOT.jar in the docker container when we are building the docker image. That way when we call spark, we can always use the same jar path every release. Then the test image always calls the same jar as well and there are fewer moving parts.  Also our customers wouldn't have to update the command (or wild card it) they use to call the jar each release.

dkwasny-ibm:
> I'm wondering if a better approach would be to rename the cohort-evaluator-spark--SNAPSHOT.jar in the docker container when we are building the docker image. That way when we call spark, we can always use the same jar path every release. Then the test image always calls the same jar as well and there are fewer moving parts. Also our customers wouldn't have to update the command (or wild card it) they use to call the jar each release.

As much as I'm ashamed to admit it, this feels like the least messy solution.
There might be _some_ value in storing the version number in a file if it's not found anywhere else in the image, but only for informational purposes.

deekim (approved):
I like Tim's idea of just making the jar name static, but I don't see anything wrong with the proposed pattern in the IBM toolchain universe.

msargentibm:
Implemented Tim's suggestion. Ran the changes through the CI/CD toolchains successfully.

deekim (approved):

tabishop (approved):
LGTM, thanks for turning the crank again and sorry for the churn

msargentibm:
@tabishop -- Thanks for the suggestion, it made the most sense. The only "hard" part was getting the CI/CD toolchains to actually run :)

dkwasny-ibm (approved):

PR #231, opened 2022-01-12
Wffhcohort 679 - add COS as possible storage location in fhir-resource-tooling

shelleyMurphy:
Adding option to write directly to COS, as well as or instead of the local file system.

### DIFF
--- a/docs/user-guide/fhir-resource-tooling.md
+++ b/docs/user-guide/fhir-resource-tooling.md
@@ -42,3 +42,18 @@ $> java -cp fhir-resource-tooling-0.0.1-SNAPSHOT-shaded.jar com.ibm.cohort.tooli
 ```
 
 Please note, if you intend to use custom mappings those mappings must be provided to load the value set with this tool.
+
+There is also the ability to write the json representation of the value set either locally or to cos, in the case that you would like to check that it looks how you would expect. In this case, rather than the server configuration, you would need to specify either COS, LOCAL, or BOTH with the --output-locations flag.
+
+
+#####Parameters
+
+* -m/--measure-server: The path to the JSON configuration data for the FHIR server connection that will be used to retrieve measure and library resources. If this is provided, `output-locations` must be **NONE**.
+* --override-existing-value-sets: If provided, will force the creation of valuesets, even if they already exist.
+* -c/--code-system-mappings: A file containing custom code system mappings to be used.
+* --output-locations: Determines if and where to write a FHIR representation of the valueset. If set to **NONE**, the value sets will be written to the specified FHIR database. If **LOCAL**, the program will use the `file-system-output-path` option to determine where to write the FHIR representation locally. If **COS**, the program will use the `cos-configuration` option to determine the specific COS instance and `bucket` to determine which bucket to write it to. If **BOTH**, the program will write the results to both places. If this is not **NONE**, the `measure-server` configuration must not be specified.
+* -p/--file-system-output-path: The local filesystem path to write results out to (will only be used if output-locations is either **BOTH** or **LOCAL**).
+* -b/--bucket: COS bucket to write results to (will only be used if output-locations is either **BOTH** or **COS**).
+* --cos-configuration: A json file containing all the relevant cos configuration needs for access
+* -o/--file-system-output-format: The format to use when exporting value sets to the file system when using the -p/--file-system-output-path parameters. Valid values are JSON or XML. If not specified, the default output format will be JSON
+* [no option]: The last argument in the command should be a list of all the spreadsheets you intend to upload.
--- a/docs/user-guide/fhir-resource-tooling.md
+++ b/docs/user-guide/fhir-resource-tooling.md
@@ -48,12 +48,22 @@ There is also the ability to write the json representation of the value set eith
 
 #####Parameters
 
-* -m/--measure-server: The path to the JSON configuration data for the FHIR server connection that will be used to retrieve measure and library resources. If this is provided, `output-locations` must be **NONE**.
-* --override-existing-value-sets: If provided, will force the creation of valuesets, even if they already exist.
-* -c/--code-system-mappings: A file containing custom code system mappings to be used.
-* --output-locations: Determines if and where to write a FHIR representation of the valueset. If set to **NONE**, the value sets will be written to the specified FHIR database. If **LOCAL**, the program will use the `file-system-output-path` option to determine where to write the FHIR representation locally. If **COS**, the program will use the `cos-configuration` option to determine the specific COS instance and `bucket` to determine which bucket to write it to. If **BOTH**, the program will write the results to both places. If this is not **NONE**, the `measure-server` configuration must not be specified.
-* -p/--file-system-output-path: The local filesystem path to write results out to (will only be used if output-locations is either **BOTH** or **LOCAL**).
-* -b/--bucket: COS bucket to write results to (will only be used if output-locations is either **BOTH** or **COS**).
-* --cos-configuration: A json file containing all the relevant cos configuration needs for access
-* -o/--file-system-output-format: The format to use when exporting value sets to the file system when using the -p/--file-system-output-path parameters. Valid values are JSON or XML. If not specified, the default output format will be JSON
-* [no option]: The last argument in the command should be a list of all the spreadsheets you intend to upload.
+```
+-m,--measure-server
+The path to the JSON configuration data for the FHIR server connection that will be used to retrieve measure and library resources. If this is provided, `output-locations` must be **NONE**.
+--override-existing-value-sets
+If provided, will force the creation of valuesets, even if they already exist.
+-c,--code-system-mappings 
+A file containing custom code system mappings to be used.
+--output-locations
+Determines if and where to write a FHIR representation of the valueset. If set to **NONE**, the value sets will be written to the specified FHIR database. If **LOCAL**, the program will use the `file-system-output-path` option to determine where to write the FHIR representation locally. If **COS**, the program will use the `cos-configuration` option to determine the specific COS instance and `bucket` to determine which bucket to write it to. If **BOTH**, the program will write the results to both places. If this is not **NONE**, the `measure-server` configuration must not be specified.
+-p,--file-system-output-path
+The local filesystem path to write results out to (will only be used if output-locations is either **BOTH** or **LOCAL**).
+-b,--bucket
+COS bucket to write results to (will only be used if output-locations is either **BOTH** or **COS**).
+--cos-configuration
+A json file containing all the relevant cos configuration needs for access
+-o,--file-system-output-format
+The format to use when exporting value sets to the file system when using the -p/--file-system-output-path parameters. Valid values are JSON or XML. If not specified, the default output format will be JSON
+```
+The last argument in the command should be a list of all the spreadsheets you intend to upload.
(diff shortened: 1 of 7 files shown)


fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -50,10 +63,19 @@
 
 		@Parameter(names = {"-h", "--help"}, description = "Show this help", help = true)
 		boolean isDisplayHelp;
-		
-		@Parameter(names = {"-p", "--file-system-output-path"}, description = "If specified, value sets will be converted to FHIR format and writen out to the file system path specified by this parameter. Value sets will not be imported into the FHIR server if this option is set.", required = false)
+
+		@Parameter(names = {"--output-locations"}, description = "If any but \"NONE\" is specified, value sets will be converted to FHIR format and written to a location specified by the bucket or filePath parameters. Value sets will not be imported into the FHIR server if this option is set. Options are COS, LOCAL, BOTH, NONE.")
+		String fileOutputLocation = "NONE";
```
deekim, line 68: I think using an `Enum` here might work out better for you.
→ [EnumConverter.java](https://github.com/cbeust/jcommander/blob/master/src/main/java/com/beust/jcommander/converters/EnumConverter.java)

I believe that the default `--help` also includes the list of `Enum` options for you, so you wouldn't have to maintain it in the `description` field.

deekim (approved):
LGTM otherwise.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -138,8 +168,29 @@ static void runWithArgs(String[] args, PrintStream out) throws IOException {
...
+		ObjectMetadata metadata = new ObjectMetadata();
+		byte[] arr = fhirContext.newJsonParser().encodeResourceToString(vs).getBytes();
+		metadata.setContentLength(arr.length);
+		PutObjectRequest put = new PutObjectRequest(arguments.bucket, vsFileName, new ByteArrayInputStream(arr), metadata);
+		cosClient.putObject(put);
+	}
+
+	public AmazonS3 createClient(String api_key, String service_instance_id, String endpoint_url, String location)
+	{
+		SDKGlobalConfiguration.IAM_ENDPOINT = "https://iam.cloud.ibm.com/identity/token";
```
tabishop, line 181: Should we be hardcoding this? Seems like this would be tying us to IBM cloud and what if this endpoint changed?

fhir-resource-tooling/src/test/resources/example-cos-credentials.json
```diff
@@ -0,0 +1,6 @@
+{
+  "apikey": "<apikey from COS service credentials>",
```
tabishop, line 2: Are we tied to IBM Cloud Object store with this? I would think we would want to use the more generic hmac keys (access key and secret key) instead of IBM Cloud specific apikeys

fhir-resource-tooling/pom.xml
```diff
@@ -72,6 +72,27 @@
 			<artifactId>slf4j-api</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>com.ibm.cos</groupId>
+			<artifactId> ibm-cos-java-sdk-core</artifactId>
```
tabishop, line 77: Is this implementation tied specifically to IBM COS? I think we would want to use a generic S3 client to allow us to write to non-IBM object stores, especially with all the upcoming talk of supporting azure.

tabishop (commented):
In general, I have concerns we are tying ourselves specifically to IBM COS with this instead of using a generic S3 implementation? Please correct me if I'm wrong.

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
shelleyMurphy: Done!

fhir-resource-tooling/src/test/resources/example-cos-credentials.json
shelleyMurphy: Updated to use HMAC

fhir-resource-tooling/pom.xml
shelleyMurphy: Yeah, fair, I've updated the dependencies.

tabishop (approved):
LGTM

jaigoradia (approved):

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -62,17 +84,20 @@
 		
 		public void validate() {
 			//check to make sure we are either exporting to file OR importing to a fhir server, but not both
-			if(fileSystemOutputPath != null && measureServerConfigFile != null) {
-				throw new IllegalArgumentException("Parameters [-m, --measure-server] and [-p, --file-system-output-path] cannot both be specified on the same invocation");
+			if(!(fileOutputLocation == OutputLocations.NONE) && measureServerConfigFile != null) {
```
msargentibm, line 87: Is this just the same as `fileOutputLocation != OutputLocations.NONE`? Or am I misreading this line?

msargentibm (approved):

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
shelleyMurphy: yeah, leftover from when it was a different compare, thanks!

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -19,23 +20,35 @@
...
 
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.rest.client.api.IGenericClient;
 
+
 public class ValueSetImporter {
 	private static final Logger logger = LoggerFactory.getLogger(ValueSetImporter.class.getName());
 	
 	private enum FileFormat {JSON, XML};
+	private enum OutputLocations{NONE, LOCAL, COS, BOTH}
```
csandersdev, line 51:
```suggestion
	private enum OutputLocations{NONE, LOCAL, S3, BOTH}
```

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/cos/CosConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+package com.ibm.cohort.tooling.cos;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+public class CosConfiguration {
```
csandersdev, line 18:
```suggestion
public class S3Configuration {
```

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -50,10 +63,19 @@
...
+		OutputLocations fileOutputLocation = OutputLocations.NONE;
+
+		@Parameter(names = {"-p", "--file-system-output-path"}, description = "Local filesystem path to write results out to (will only be used if output-locations is either BOTH or LOCAL.")
 		String fileSystemOutputPath;
-		
+
+		@Parameter(names = {"-b", "--bucket"}, description = "Bucket to write results out to (will only be used if output-locations is either BOTH or COS).")
+		String bucket;
+
+		@Parameter(names = {"--cos-configuration"}, description = "a json file containing all the relevant cos configuration needs for access")
```
csandersdev, line 76:
```suggestion
		@Parameter(names = {"--s3-configuration"}, description = "a json file containing all the relevant cos configuration needs for access")
```

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -50,10 +63,19 @@
...
+		File cosJsonConfigs;
```
csandersdev, line 77:
```suggestion
		File s3JsonConfigs;
```

fhir-resource-tooling/src/main/java/com/ibm/cohort/tooling/fhir/ValueSetImporter.java
```diff
@@ -138,8 +171,28 @@ static void runWithArgs(String[] args, PrintStream out) throws IOException {
 		}
 	}
 
+	private static void putToCos(ValueSetImporterArguments arguments, FhirContext fhirContext, ValueSet vs, String vsFileName, AmazonS3 cosClient) {
```
csandersdev, line 174:
```suggestion
	private static void putToS3(ValueSetImporterArguments arguments, FhirContext fhirContext, ValueSet vs, String vsFileName, AmazonS3 s3Client) {
```

fhir-resource-tooling/src/test/resources/example-cos-credentials.json
```diff
@@ -0,0 +1,6 @@
+{
+  "access_key_id": "<HMAC access_key_id from COS service credentials>",
+  "secret_access_key": "<HMAC secret_access_key from COS service credentials>",
```
csandersdev, line 3: Did you think about whether or not this variable should be encrypted within the config file? It used to be that passwords were required to be encrypted in config files. With K8S and secrets, I am much less clear on the security requirements.

docs/user-guide/fhir-resource-tooling.md
```diff
@@ -42,3 +42,18 @@ $> java -cp fhir-resource-tooling-0.0.1-SNAPSHOT-shaded.jar com.ibm.cohort.tooli
 ```
 
 Please note, if you intend to use custom mappings those mappings must be provided to load the value set with this tool.
+
+There is also the ability to write the json representation of the value set either locally or to cos, in the case that you would like to check that it looks how you would expect. In this case, rather than the server configuration, you would need to specify either COS, LOCAL, or BOTH with the --output-locations flag.
+
+
+#####Parameters
+
+* -m/--measure-server: The path to the JSON configuration data for the FHIR server connection that will be used to retrieve measure and library resources. If this is provided, `output-locations` must be **NONE**.
```
csandersdev, line 51: Pretty much everywhere else in the user guide, if we are displaying the program usage, we use a triple-backtick pre-formatted box and just copy/paste in the program usage. Not sure this is a hard request to change since you are using some bold formatting, but inconsistency bothers me.

csandersdev (commented):
Left some fairly minor feedback and questions. Generally, I am ok with the changes, but think that maybe we update names to reflect the COS -> S3 change that Tim requested.

I know it wasn't part of the ticket, but I'm also wondering about whether or not we should add local-only and COS storage to the measure importer as well. The features exposed by the CLIs are not consistent anymore.

PR #233, opened 2022-01-13
Wffhcohort 707 - Update CLI interfaces to use command pattern 

msargentibm:
This PR changes the entrypoints in the `cohort-cli` and `fhir-resource-tooling` modules so that using the `-jar` command allows you to access multiple CLIs behind the scenes.

Available commands for `cohort-cli` module:
* `cohort-cli` -> runs CohortCLI
* `measure-cli` -> runs MeasureCLI
* `translation-cli` -> runs TranslationCLI

Available commands for `fhir-resource-tooling` module:
* `measure-importer` -> runs MeasureImporter
* `value-set-importer` -> runs ValueSetImporter

Example command:
` java -jar fhir-resource-tooling/target/fhir-resource-tooling-2.0.1-SNAPSHOT-shaded.jar measure-importer <other options for measure importer>`

I also noticed the TranslationCLI had some extraneous options related to a FHIR server connection. I removed those for now, but can add them back if folks want.


jaigoradia (commented):
Matt.  Looking at the interface changes i think the current fvt tests for cohortcli and measurecli that i have will probably fail. Could you make a couple of minor changes for me as part of this PR to the following two files: I would really appreciate it. Thanks. 
1. In .../tests/src/main/python/TestDriver_CohortCLI.py  edit line#37 to add cohort-cli in the callDetails like so: 
callDetails = ["java", "-Xms1G", "-Xmx1G", "-Djavax.net.ssl.trustStore="+os.environ["TRUSTSTORE"], "-Djavax.net.ssl.trustStorePassword="+os.environ["TRUSTSTORE_PASSWORD"], "-Djavax.net.ssl.trustStoreType="+os.environ["TRUSTSTORE_TYPE"], "-jar", jar, "cohort-cli", "-f", libraries, "-l", library]

2. Similary edit file .../tests/src/main/python/TestDriver_MeasureCLI.py and change the callDetails there line#40 replacing the "com.ibm.cohort.cli.MeasureCLI" at the very end with just "measure-cli".   

I think that should make the tests work from what i can tell. Thanks again.

csandersdev (commented):
I'm pretty happy with all of this, but it looks like all of the new files are missing copyrights.

msargentibm:
@CoreySandersOH -- Doh. Thanks! I can fix that

tabishop (approved):
Looks good to me. Nit: Would it make sense to add a small note to the docs indicating where the API has changed from previous releases, in case it breaks anyone so they know what changed?

msargentibm:
> Nit: Would it make sense to add a small note to the docs indicating where the API has changed from previous releases, in case it breaks anyone so they know what changed?

This is a good idea. Do we have release notes or something similar when we cut a release? We could always post potentially breaking changes there.

tabishop:
> > Nit: Would it make sense to add a small note to the docs indicating where the API has changed from previous releases, in case it breaks anyone so they know what changed?
> 
> This is a good idea. Do we have release notes or something similar when we cut a release? We could always post potentially breaking changes there.

I don't think we've been tracking stuff for release notes, but I think it would make sense to open a ticket and add stuff to it throughout the release and then create a release notes file at the end of the release.

csandersdev:
> I don't think we've been tracking stuff for release notes, but I think it would make sense to open a ticket and add stuff to it throughout the release and then create a release notes file at the end of the release.

Based solely on the projects I interact with (DBCG stuff, IBM FHIR), they add the documentation for the release on the release page in Github (e.g. https://github.com/Alvearie/quality-measure-and-cohort-service/releases/tag/v2.0.1). I'm not sure how that gets created, but it looks like we can go back and edit as needed. I think that is where we should also keep our notes.

jaigoradia (approved):

PR #236, opened 2022-01-25
Wffhcohort 491 spark history log cleanup

tabishop:
Pass in the parameters needed to enable spark history server log cleanup.  SPARK_HISTORY_OPTS was already being set in the entrypoint.sh file called by the history server Dockerfile, so just needed to set the parms and pass them in via the helm chart.

### DIFF
--- a/chart/cohort/templates/spark-history-server/deployment.yaml
+++ b/chart/cohort/templates/spark-history-server/deployment.yaml
@@ -49,6 +49,14 @@ spec:
           value: {{ .Values.sparkHistoryServer.s3.bucket }}
         - name: AWS_LOG_PATH
           value: {{ .Values.sparkHistoryServer.s3.eventsDir }}
+        - name: SPARK_HISTORY_FS_CLEANER_MAXAGE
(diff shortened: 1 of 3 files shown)


jillrdoty (approved):

chart/cohort/templates/spark-history-server/deployment.yaml
```diff
@@ -49,6 +49,8 @@ spec:
           value: {{ .Values.sparkHistoryServer.s3.bucket }}
         - name: AWS_LOG_PATH
           value: {{ .Values.sparkHistoryServer.s3.eventsDir }}
+        - name: SPARK_HISTORY_OPTS
+          value: {{ .Values.sparkHistoryServer.logs.sparkHistoryOpts }}
```
dkwasny-ibm, line 53: Pedantic:  I don't really see the need for the `logs` part of the identifier hierarchy.
I would be fine with just `sparkHistoryServer.sparkHistoryOps`.

dkwasny-ibm (approved):
Nothing worth blocking over.

PR #245, opened 2022-02-18
Fix Sonar NPEs

deekim:
_resolves_ [Sonar Bugs](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=Alvearie_quality-measure-and-cohort-service) identified in latest scan


cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/aggregation/ContextRetriever.java
```diff
@@ -312,9 +313,13 @@ public ContextRetriever(Map<String, String> inputPaths, DatasetRetriever dataset
         if (path == null) {
             throw new IllegalArgumentException(String.format("No path mapping found for datatype %s", dataType));
         }
+
         Dataset<Row> dataset = datasetRetriever.readDataset(path);
-        return ( dataset != null ) ? dataset
-                .withColumn(SOURCE_FACT_IDX, functions.lit(dataType)) : null;
+        if (dataset == null) {
+            throw new IllegalArgumentException(String.format("Could not read data for %s", dataType));
+        }
```
dkwasny-ibm, line 320: I believe @csandersdev might have something to say about this.  I remember the original implementation of this class being null hostile, but being changed later to its current state.

msargentibm (approved):
Changes make sense as long as Kwas' comment is addressed.

cohort-evaluator-spark/src/main/java/com/ibm/cohort/cql/spark/aggregation/ContextRetriever.java
deekim: For what it's worth, the method's scope is `private` and a similar `Exception` is already being throwing as a precondition check in the method:
```java
if (path == null) {
  throw new IllegalArgumentException(String.format("No path mapping found for datatype %s", dataType));
}
```

Any additional feelings on this @csandersdev?

shelleyMurphy (approved):

PR #246, opened 2022-02-28
newly regenerated code from swagger jar that includes api client code for enhanced health check REST endpoint.

jaigoradia:

Note this PR is to merge swagger auto-generated code for REST API endpoints.  I had to regenerate the code so that it would include the api client glue code for the new enhanced health check REST endpoint.  Since this is auto generated code by a tool there is no need to review the files included as part of this PR. I will have another PR coming shortly that includes the test files i added/modified to test the new endpoint. Did not want to mix the two sets of files since these auto generated files can just be given a rubber stamp approval.


tabishop (approved):
LGTM

PR #247, opened 2022-02-28
New and updated test files in support of testing the enhanced health check REST API endpoint.

jaigoradia:

This PR contains a mix of new and updated test files for the following two scenarios:
1. Testing the new enhanced health check REST API endpoint.  Both positive and negative testing is being done.
2. In addition to adding new automated tests for the new enhanced health check REST API endpoint i am also piggy backing on this PR to remove usage of persistent volume claim references to our COS buckets from spark-submit. For example i removed one of the two spark tests that was being called from run-fvttests.sh since that one was using a spark-submit with direct reference to the config files in cohort-config COS bucket instead of making s3a:// calls to the same files. I am doing this in anticipation of Tim working on a separate ticket to remove all linkages for PVCs being created via Terraform and elsewhere.  By the same token i also removed references to PVCs for cohort-config and cohort-data-tenant2 COS buckets from the spark-defaults.conf file that the spark tests use for fvt.

Note: While investigating how i could get rid of all references for PVCs from the code i did notice that the test pod that i spin off to run spark tests in, does mount a storage volume backed by a PVC for the cohort-data-tenant2 COS bucket so that in the run-fvttests.sh script i am then able to copy over the  output files of the spark test from the COS bucket and make them available locally to the python validation script that validates the output for accuracy within the toolchain environment.  
For this reason i do have a dependency on this one PVC being there in the test environment.  In order to address Tim's concern about the possibility of that PVC not being there in say Stable namespace like he ran into, i added some automation to the create-sparkfvt-pod.sh script to ensure that if the PVC did not exist it would automatically get created before the pod is spun off.


tabishop (approved):
Looks good. Thanks for adding the stuff to check for the PVC and create it if it was not present.

PR #257, opened 2022-03-18
Test CodeQL fork trigger

deekim:


deekim:
Verified that the [CodeQL workflow](https://github.com/Alvearie/quality-measure-and-cohort-service/actions/runs/2004548950) gets triggered when PR is opened from fork.

PR #260, opened 2022-03-24
Add custom context name support

deekim:
_Defaulting_ to `Patient` context to preserve backward compatibility

## todo
- [x] _address_ [Corey's feedback](https://github.com/Alvearie/quality-measure-and-cohort-service/pull/260#pullrequestreview-924748416) on adding documentation

## References
- [WFFHCOHORT-861](https://jira.wh-sdlc.watson-health.ibm.com/browse/WFFHCOHORT-861)


jillrdoty (approved):
I know it's draft but LGTM

csandersdev (commented):
I did some testing of alternate (e.g. not Patient) context names for our consulting team client and I found that my unit tests worked in places where the actual runtime failed. The core issue was related to the magic define that gets added to every CQL library (e.g. define Encounter: singleton from [Encounter] e). If the context specified isn't a valid FHIR resource, and no expression filter is present, then that define will puke. You've got it under control in your unit test, but I feel like there is a documentation note that could/should show up in the user docs (speaking of which, you should add this new parameter to the user guide). 

The context names that are valid but are not FHIR resources are `Population` and `Unspecified`. If you try to use any other name that isn't a valid FHIR resource, the FhirModelResolver will blow up. It might be worthwhile to include something along those lines in the parameter description or maybe just in the user guide when you make those changes.

msargentibm:
@deekim -- I do not have a model in mind. I just saw that there is a model-info argument in the usage block and didn't remember that CohortCLI is FHIR based only.

csandersdev:
> @csandersdev would you happen to know?

I don't have a firm answer. There is a contextInfo node in the modelinfo, but I'm not sure whether the translator _requires_ those for a valid using statement or not. 

The Unspecified and Population names are hard-coded into the [FHIRModelResolver](https://github.com/DBCG/cql_engine/blob/43043a1a5ba9a3b615d2824fc861ec499f880a2e/engine.fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/model/FhirModelResolver.java#L93). There is also some hard-coding in the [SearchParamFhirRetrieveProvider](https://github.com/DBCG/cql_engine/blob/43043a1a5ba9a3b615d2824fc861ec499f880a2e/engine.fhir/src/main/java/org/opencds/cqf/cql/engine/fhir/retrieve/SearchParamFhirRetrieveProvider.java#L125) that will check for a context named Patient.

deekim:
> I do not have a model in mind. I just saw that there is a model-info argument in the usage block and didn't remember that CohortCLI is FHIR based only.

👍 Gotcha, I think that just if you had a custom `FHIR`-based `IG`, but I could be wrong there.

PR #272, opened 2022-05-17
Wffhcohort 822 - Support transitive joins in spark implementation

msargentibm:
This PR changes the structure of `ContextDefintion` so that related data is joined in to a context using a user-specified `joinClause`. It may be useful to take a look at the `docs` changes to get a feel for what is happening before diving into the code.

`ContextRetriever` is probably a good place to start looking through the code since most of the logic changes live there. Most other changes involve moving, deleting, or renaming code.


docs/user-guide/spark-user-guide.md
```diff
@@ -290,9 +306,33 @@ entity "**vocabulary**" {
...
+        },
+        {
+          "name": "observation",
+          "joinClause": "inner join observation on person.person_id = observation.person_id"
+        }
+      ]
+    }
+  ]
+}
+```
```
rickjstevens, line 335: Not necessarily a concern, but wanted to confirm an observation I had. It looks like each of these joinClause statements are meant to be independent of each other and all based on/relative to primaryDataType? Versus incremental and based on prior joinClause statements? Assume the code and produces the final SQL statement to use for a context is able to filter out redundant logic across multiple joinClause expressions?

msargentibm: Your observation is correct. The joins work in the same manner that they did before these changes. Each join is independent.

shelleyMurphy (approved):
Looks much cleaner, I'm a fan!

rickjstevens (approved):
Looks good - I like the added flexibility this will enable for future use of our cohort spark implementation.

PR #277, opened 2022-05-26
Updates to ossindex exclusions

rickjstevens:
Signed-off-by: Rick Stevens <rstevens@us.ibm.com>

### DIFF
--- a/cohort-parent/pom.xml
+++ b/cohort-parent/pom.xml
@@ -1432,6 +1432,14 @@
 						<excludeVulnerabilityId>sonatype-2021-4682</excludeVulnerabilityId>
 						<excludeVulnerabilityId>sonatype-2020-0926</excludeVulnerabilityId>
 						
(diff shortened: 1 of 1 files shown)


rickjstevens:
I had to make more extensive changes to exclusion rules for sonatype/ossindex vulnerabilities. In adding to excluding latest items being flagged, I had to update all prior exclusion rules given ossindex changed their vulnerability id model so that our prior rules would no longer work.

shelleyMurphy (approved):
Can we delete the commented out exclusions? I don't know if we'll need to reference the old IDs in the future, but I hope not.

rickjstevens:
> Can we delete the commented out exclusions? I don't know if we'll need to reference the old IDs in the future, but I hope not.

Agree - I suppose we can always look at commit history if we ever need those old ids again - pretty doubtful they will be needed going forward. Will remove them.

dkwasny-ibm (approved):

msargentibm (approved):

